PR #8, opened 2015-03-08
Code style according PEP8 - 'Style Guide for Python Code'. Code format.

Saylenty:
Have a look at the document: https://www.python.org/dev/peps/pep-0008/
Additionally, some redundant parentheses, semicolons, unused variables, imports were removed.
Simplified some expressions in 'if', 'while' loops, etc.


Saylenty:
Dear Kbandla,

I can't understand why my commit is worse? I have compared your code and 
my on https://landscape.io and github, and my local machine. But problem 
seems to be  that some files aren't checked by landscape.io (e.g. 
dpkt/__init__.py) whereas my commit was entirely processed.
Might you please clarify the situation?
Thanks

Sincerely,
Alex

brifordwylie:
@Saylenty Thanks for the PR :)  The code has a long history starting from the original project https://code.google.com/p/dpkt/. There is obviously lots of room for cleanup, in general we're trying to balance that cleanup with avoiding changes that might negatively impact existing users. In particular the @property changes are great (thumbs up) but if an outside script is calling get_opcode() for instance and that method is now gone than we might upset some folks. So the PR is great but we might want to include any old getters and setters with a small wrapper that issues a DeprecationWarning and then calls the new code underneath. We'll do that for a month or two and then remove them. Sound reasonable?

brifordwylie:
@Saylenty This looks great (doing it for the _ methods may have been overkill but that's okay :). You can see that Travis is failing, you can run 'tox' in the top level directory to replicate the failure, the problem is that since the configs/ directory is not in the package it can't be properly located. The deprecated decorators are super nice so lets go ahead and put them into the dpkt pacakge proper. I'm going to merge in your changes and then move the decorator code.

Saylenty:
@brifordwylie I had such problem before committing the code. These problems with locating the package were resolved by adding some extra lines into "dpkt/dpkt.egg-info/SOURCES.txt" file. These lines are:
configs/(2 underscores)init(2 underscores).py
configs/decorators.py
Then I deleted *.pyc files and ran "tox" again and all tests passed successfully.
I thought that this work (decorator) has to be in the separate package because in the "dpkt" package all staff is used by users whereas decorator shouldn't. Package name "configs" is not very suitable because this is not configurable option or staff, but I couldn't come up a better name :)

PR #10, opened 2015-03-22
deprecated_method_decorator fix.

Saylenty:
The decorator don't show the actual method name should be called instead.
Avoid potential mistakes.

### DIFF
--- a/dpkt/decorators.py
+++ b/dpkt/decorators.py
@@ -5,15 +5,8 @@ import warnings
 def deprecated_method_decorator(deprecated_method):
     def wrapper(*args, **kwargs):
         warnings.simplefilter('always', DeprecationWarning)  # turn off filter
-        if str(deprecated_method.__name__).startswith("_get_"):  # getter method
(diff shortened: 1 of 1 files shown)


brifordwylie:
okay, it's fine if you do some more/others. I put in some new tests on https://github.com/kbandla/dpkt/pull/11. The performance is definitely an issue as @Saylenty alluded to all those print outs kill performance. So... look.. my take is that the message is like a big red sign 'Change your code!'.. and when people do the performance will be back to normal :)

I'd rather keep the 'Big Red Sign', again the bad performance will disappear when they change their code and that's exactly what we want to encourage.

Saylenty:
I'm a bit scared and don't know might I create a new pull request or not, so could you look at the code and suggest can I or not :)?

brifordwylie:
I think the code is fine right now. After the #11 merge we have two tests against the deprecated decorators and one test for the performance. So do a PR if you want to but you don't need to at this point. Also just in general doing a PR is good/fine as it won't have any negative consequences until it's merged and @kbandla or I will look at it before the merge.

PR #11, opened 2015-03-24
converting a few tests to pytest 'style' and testing deprecation decorators

brifordwylie:


brifordwylie:
The code changes look like way more than they are. I've simply move the test code, not created or removed it.

brifordwylie:
You're the best, have at it. :)

PR #12, opened 2015-03-25
deprecated_method_decorator supplementation

Saylenty:
unittest + one message per one method call regardless of location


brifordwylie:
@Saylenty we're using pytest now, which simply automatically detects any method with test* in the name and runs it. IMHO much nicer and 'flatter' than unittest. So look at http.py or dns.py for an example.. basically just remove all unitest boilerplate,

brifordwylie:
oh and good thinking on the 'once' flag that will obviously be nicer than spamming with messages :)

brifordwylie:
I'll go ahead and merge this with the expectation that the test will be fixed in a later PR.

Issue #17, opened 2015-05-18
IP Flags and Fragment Offset


brifordwylie:
Right now dpkt simply shoves these 16 bits into the same field ip.off, so at a minimum we should have some documentation and an example showing how to separate the two...

    do_not_fragment = self.packet.off & dpkt.ip.IP_DF
    more_fragments = self.packet.off & dpkt.ip.IP_MF
    fragment_offset = self.packet.off & dpkt.ip.IP_OFFMASK

brifordwylie:
Closed by PR #19

PR #19, opened 2015-05-21
adding some much needed example code

brifordwylie:


brifordwylie:
Closes: IP Flags and Fragment Offset #17

Issue #15, opened 2015-03-31
Landscape Error Cleanup


brifordwylie:
https://landscape.io/github/kbandla/dpkt/29/messages/error

Someone just needs to go through these and figure out what's going on, if real error fix, if not real error put in a comment why it's fine.

kbandla:
This is not a real error. The property in question is populated at runtime by the base packet class based on the value of __hdr__.

brifordwylie:
Ah okay, so if all 11 errors are 'mis-marked' than I'll just close this issue.

kbandla:
Yes, all those 11 are non-errors. We can go ahead and close this issue. 

We could 'fix' these [errors](https://github.com/kbandla/dpkt/blob/master/dpkt/icmp.py#L116-L117) by doing something like this:
```
if not hasattr(self, 'sum'):
    self.sum = dpkt.in_cksum(dpkt.Packet.__str__(self))
```
However, `hasattr` (and `getatrr`) internally use `try/except`. So its faster to use 'try/except' instead. But again, all this just to make the linter happy. This is a non-issue at runtime..

brifordwylie:
Closing Issue

PR #24, opened 2015-05-27
Better testing for 'duration decorator' + code format + 'FIXME' marks.

Saylenty:


Saylenty:
It seems to me that such kind of errors like 'Access to member xxx before its definition line yyy' are mistaken. However, these ones should be commented at least. What are your views?

kbandla:
This looks like a duplicate of issue #15. The properties are populated at runtime by the base packet class.

Saylenty:
@kbandla Thanks for explanation. You can discard the pull request.

PR #26, opened 2015-05-31
add missing conversions to str in __str__()

richardhansen:
Don't assume that the raw packet data is a `str` object; convert it to `str` via the `str()` function first.  (The raw packet data might be a string-like buffer object.)

### DIFF
--- a/dpkt/ip.py
+++ b/dpkt/ip.py
@@ -63,7 +63,7 @@ class IP(dpkt.Packet):
 
     def __str__(self):
         if self.sum == 0:
-            self.sum = dpkt.in_cksum(self.pack_hdr() + self.opts)
+            self.sum = dpkt.in_cksum(self.pack_hdr() + str(self.opts))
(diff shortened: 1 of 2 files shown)


sunhao2014:
The update is definitely correct. It is more reasonable using `str()` function to the `self.opts`. Also I checked the project for similar problems. It seems that only `tcp.py` and `ip.py` have this problem.

However I have a question. If the raw input is a `buffer` object, the original implementation also works, for it'll convert the input to `str` implicitly. Thus do we have any other kind of raw inputs? If not, the original implementation would also be correct.

richardhansen:
> Thus do we have any other kind of raw inputs?

Yes.  A project I was working on put packet data in a custom C-based object type that implemented the [buffer protocol](https://docs.python.org/2/c-api/buffer.html).  (It had a `__str__()` method to convert the data to a string, but the add operator couldn't be used to concatenate the data to another string.)

Thank you for merging!

sunhao2014:
@richardhansen. Got it! Thanks for sharing us the information.

PR #27, opened 2015-05-31
fix bug in IPv6 payload checksum generation

richardhansen:
When determining whether a checksum must be generated for the payload look at `IPV6.p` not `IPV6.nxt`.  `IPV6.p` is the protocol number of the payload, while `IPV6.nxt` is the protocol number of the next extension header (or payload if there is no extension header).


brifordwylie:
look good, merging.

richardhansen:
Thank you!

Issue #117, opened 2015-06-04
dpkt http wrong parse_headers logic


kbandla:
_From [almog1...@gmail.com](https://code.google.com/u/106388495642043329325/) on May 21, 2012 05:52:27_

I have been encountered a problem which I found also some people complaining about on the net.

The description on the error from here http://stackoverflow.com/questions/6337878/parsing-pcap-files-with-dpkt-python is about the dpkt.http.parse_headers function
"""
While it seems to effectively parse most of the packets, I'm receiving a NeedData("premature end of headers") exception on some. They appear to be valid packets within WireShark, so I'm a bit confused as to why the exceptions are being thrown.
"""

I investigated the problem and found that the exception is raised if the HTTP is not ending with "\r\n\r\n".

I have researched more than 10 Milion Http Request and there wasn't even one packet where this exception was relevant.

I changed the parse headers on my installatioin to look like this

def parse_headers(f):
    """Return dict of HTTP headers parsed from a file object."""
    d = {}
    while 1:
        line = f.readline()
        if not line:
            break
        \# I commented  this line out because the way this parser works is wrong.
        \# When the http does not finish with '\r\n\r\n' the parsing is faild!
        \#    raise dpkt.NeedData('premature end of headers')

        \# The next two liness are responsibble for stopping the parsing when the
        \# line equals '\r\n'
        line = line.strip()
        if not line:
            break
        l = line.split(None, 1)
        if not l[0].endswith(':'):
            raise dpkt.UnpackError('invalid header: &#37;r' &#37; line)
        k = l[0][:-1].lower()
        v = len(l) != 1 and l[1] or ''
        if k in d:
            if not type(d[k]) is list:
                d[k] = [d[k]]
            d[k].append(v)
        else:
            d[k] = v
    return d

and its working great.

Many thanks,
Almog

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=90_

kbandla:
_From [jeffrey.guy@gmail.com](https://code.google.com/u/jeffrey.guy@gmail.com/) on August 01, 2012 14:56:08_

There's a separate but related bug here: l = line.split(None, 1)

The HTTP RFC indicates the delim character between headers and values is ":".  dpkt current and this patch assume there's a space following the header name.

i.e., 
Header:Value   
vs. 
Header: Value

Both are valid HTTP headers.  That block is more appropriately:

 15         line = line.strip()
 16         if not line:
 17             break
 18         l = line.split(":", 1)
[...]

PR #181, opened 2015-06-14
Fix Issue #117: dpkt http wrong parse_headers logic.

sunhao2014:
Hi @kbandla, I did some investigations on Issue #117. Here is a detailed description for this bug.

The problem is dpke could parse most of the HTTP packets, however sometimes it may raise `NeedData("premature end of headers")` exception. 

Hopefully the HTTP header ends with `\r\n\r\n`, but practically some of the packets are not with that ending. Here is an example I captured using Wireshark.

`Host: x.jd.com
Connection: keep-alive
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) 
 Chrome/40.0.2214.115 Safari/537.36
Referer: http://y0.ifengimg.com/mappa/2015/01/22/8467fe372da309ca74727a4af91ee20c.html
Accept-Encoding: gzip, deflate, sdch
Accept-Language: zh-CN,zh;q=0.8,en;q=0.6
Cookie: _pst=sunhao2014; TrackID=1PWdcr3MO_C611BGWJi_u-nDotO7jC8xU2M_oAON-
 W58NCds3C1F6bQRYDFFeSeMydoRk5R2qo8DJFNMIii1uoA; pinId=e3nZh6DHdRzvH6PAy6UGog; 
 pin=sunhao2014; unick=DreamingInCode; _tp=gKQwy5c16Th%2BRWjdagTF4g%3D%3D; user-
 key=95f5a8ba-8e80-4ebf-8855-8405a9195678; __utmz=122270672.1433142066.1.1.utmcsr=trade.jd.
 com|utmccn=(referral)|utmcmd=referral|utmcct=/shopping/order/getOrderInfo.action; 
 unpl=V2_ZzNsbUZeREd0DBRccxtfAGJRQF9KVEccc11DVXkRC1UzAhtaclRCFXEUR1NnGV8UZgoZWUN
 cQBdFCHZXfB9sBmQHFl1yZ0MWRQl2VHMZXQBvAhFaSmc%3D; mt_subsite=|125%2C1434070223|; 
 cn=0; ads_info=t=2...`

In our current implementation, we'll raise the aforementioned `NeedData` exception. However we can see the attached HTTP packet is valid and should be parsed without exception. Thus I modified the original logic. There are two cases which will lead to the `break` clause on Line 16. One is we have reached the end of header string but it doesn't end with '\r\n\r\n', for this case, we directly complete the header parsing with out raising the exception. The other case is the header part ends, so after `strip()` we'll also break the loop and continue parsing the body.

### DIFF
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -11,8 +11,6 @@ def parse_headers(f):
     d = {}
     while 1:
         line = f.readline()
-        if not line:
-            raise dpkt.NeedData('premature end of headers')
         line = line.strip()
         if not line:
             break


dpkt/http.py
```diff
@@ -11,8 +11,6 @@ def parse_headers(f):
     d = {}
     while 1:
         line = f.readline()
```
brifordwylie: I had to stare at this code for a long time before I realized that the logic covers two cases:
1) If the line = '\r\n' the strip will remove those and give ''
2) if the file(a buffer in a filelike object) is complete then readline returns ''

Please put the information above in a comment right above 'if not line:'

brifordwylie:
Also lets put in a test for a good header, and a header with a 'weird' ending (whatever we want that to be)
``` python
    s = 'POST /main/redirect/ab/1,295,,00.html HTTP/1.0\r\nReferer: http://www.email.com/login/snap/login.jhtml\r\n' \
    'Connection: Keep-Alive\r\nUser-Agent: Mozilla/4.75 [en] (X11; U; OpenBSD 2.8 i386; Nav)\r\nHost: ltd.snap.com\r\n' \
    'Accept: image/gif, image/x-xbitmap, image/jpeg, image/pjpeg, image/png, */*\r\nAccept-Encoding: gzip\r\n' \
    'Accept-Language: en\r\nAccept-Charset: iso-8859-1,*,utf-8\r\nContent-type: application/x-www-form-urlencoded\r\n' \
    'Content-length: 61\r\n\r\n' \
    'sn=em&mn=dtest4&pw=this+is+atest&fr=true&login=Sign+in&od=www'
    r = Request(s)
    assert r.method == 'POST'
    assert r.uri == '/main/redirect/ab/1,295,,00.html'
    assert r.body == 'sn=em&mn=dtest4&pw=this+is+atest&fr=true&login=Sign+in&od=www'
    assert r.headers['content-type'] == 'application/x-www-form-urlencoded'

    s_weird_end = 'POST /main/redirect/ab/1,295,,00.html HTTP/1.0\r\nReferer: http://www.email.com/login/snap/login.jhtml\r\n' \
    'Connection: Keep-Alive\r\nUser-Agent: Mozilla/4.75 [en] (X11; U; OpenBSD 2.8 i386; Nav)\r\nHost: ltd.snap.com\r\n' \
    'Accept: image/gif, image/x-xbitmap, image/jpeg, image/pjpeg, image/png, */*\r\nAccept-Encoding: gzip\r\n' \
    'Accept-Language: en\r\nAccept-Charset: iso-8859-1,*,utf-8\r\nContent-type: application/x-www-form-urlencoded\r\n' \
    'Content-length: 61\r\n\n' \
    'sn=em&mn=dtest4&pw=this+is+atest&fr=true&login=Sign+in&od=www'
    r = Request(s_weird_end)
```

dpkt/http.py
```diff
@@ -11,8 +11,6 @@ def parse_headers(f):
...
-        if not line:
-            raise dpkt.NeedData('premature end of headers')
         line = line.strip()
```
brifordwylie: I had to stare at this code for a long time before I realized that the logic covers two cases:
1) If the line = '\r\n' the strip will remove those and give ''
2) if the file(a buffer in a filelike object) is complete then readline returns ''

Please put the information above in a comment right above 'if not line:'

brifordwylie:
Also @sunhao2014  please make branches from kbandla/dpkt. That way other collaborators can checkout the branch and make changes (for instance in this case way easier for me to make the changes in the code, test it out etc.. then try to describe the changes that I want you to make :)

sunhao2014:
Hi @brifordwylie, thanks a lot for the comments and suggestions. I've created PR #189 to realize the updates you mentioned and also created a new branch called 'hao_gsoc'.

dpkt/http.py
obormot: with this change lines 13 and 14 should be merged into one statement:
```
line = f.readline().strip()
```

sunhao2014: Good suggestion, however for this point the logic is a little bit tricky. Actually it covers two cases (please see the comment in the updated source file), and I think it would be clearer if we keep this part as two  lines of code.

obormot: The 2 lines are right next to each other in the new code. Logic-wise there's no difference between
```python
line = f.readline()
line = line.strip()
```
and
```python
line = f.readline().strip()
```
the latter is just faster and shorter.

sunhao2014: Agree. In fact I mean the two-line implementation might make the user understand the logic more easily. Definitely the one-line implementation is more elegant. I'll update this part in next PR. Thanks, Oscar.

Issue #105, opened 2015-06-04
pcap.py does not implement next() -- breaks in dispatch if cnt > 0


kbandla:
_From [timur.al...@gmail.com](https://code.google.com/u/117790548747779378711/) on October 12, 2011 12:39:02_

What steps will reproduce the problem? 1. instantiate a pcap reader
2. call dispatch with cnt > 0
3. exception that next() is not implemented What is the expected output? What do you see instead? next() method should be implemented

I attached a patch that adds the next() method. If no one sees any issues with it, I'll commit it to the repository

**Attachment:** [pcap.patch](http://code.google.com/p/dpkt/issues/detail?id=78)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=78_

PR #182, opened 2015-06-14
Fix multiple pcap.Reader issues

obormot:
- make pcap.Reader's fileobj STDIN and [c]StringIO compatible;
- fix a dispatch() issue #105 and make it closer to pypcap implementation (took the docstring from there);
- add unit tests.


brifordwylie:
looks good, merging.

PR #183, opened 2015-06-15
Fix #117 and add related test cases

### LINKED ISSUE #117: dpkt http wrong parse_headers logic

sunhao2014:
Fix #117, add comments to the key logic and add test cases provided by @brifordwylie.

### DIFF
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -10,9 +10,14 @@ def parse_headers(f):
     """Return dict of HTTP headers parsed from a file object."""
     d = {}
     while 1:
+        # The following logic covers two kinds of loop exit criteria.
+        # 1) If the header is valid, when we reached the end of the header,
+        #    f.readline() would return with '\r\n', then after strip(),
+        #    we can break the loop.
+        # 2) If this is a weird header, which do not ends with '\r\n',
+        #    f.readline() would return with '', then after strip(),
+        #    we still get an empty string, also break the loop.
         line = f.readline()
-        if not line:
-            raise dpkt.NeedData('premature end of headers')
         line = line.strip()
         if not line:
             break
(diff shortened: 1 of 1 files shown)


dpkt/http.py
```diff
@@ -247,6 +252,40 @@ def test_request_version():
     except:
         pass
 
+def test_invalid_header():
+    # valid header.
+    s = 'POST /main/redirect/ab/1,295,,00.html HTTP/1.0\r\n' \
```
kbandla: This is okay too, but you can also write long strings like:
s = ( 'aaaaaa'
        'bbbbbbb'
)

brifordwylie:
look good merging. Also lets all start using https://waffle.io/kbandla/dpkt for issue management.
1) Looking/thinking about an issue? Put it 'On Deck'
2) Working on an issue? Put it 'In Progress' and make sure you assign yourself.

The boxes can just be dragged to the different columns... it also updates/syncs with GitHub so it works pretty nice.

dpkt/http.py
sunhao2014: Yeah, hopefully for this case we'll get an exception. I'll add this case to the unit test.

sunhao2014:
Cool. I tried the waffle. It's really a great issue tracking tool.

dpkt/http.py
sunhao2014: @kbandla, done! Please see 36b81cd.

PR #187, opened 2015-06-16
add some unit tests to increase code coverage

obormot:


brifordwylie:
Looks good, merging (also @obormot fell free to merge your PRs after build/health check).

brifordwylie:
Almost a + 1% coverage increase, good job @obormot

obormot:
@brifordwylie I've noticed that __main__() function calling all the tests is always a miss on the coverage report, so the paradox is that 100% coverage is only achieved when there's no tests at all...  Zen ;-)   @@deprecated old-style property getters/setters are misses as well.

Issue #164, opened 2015-06-04
[patch] Made http.py return ordered headers


kbandla:
_From [alex...@gmail.com](https://code.google.com/u/110156607140959632921/) on March 28, 2015 16:11:26_

Not a bug, just an improvement.

HTTP headers are returned unordered (they are held in a simple dict). Simple patch to make headers be returned in the same order as they are in the request.

This is helpful for the very particular use case to fingerprint certain browsers.

This does have some drawbacks, since OrderedDicts are python 2.7 and above only. Backwards compatibility can be achieved using ordereddict, but it introduces an extra dependency.

**Attachment:** [ordered_http_headers.patch](http://code.google.com/p/dpkt/issues/detail?id=138)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=138_

PR #188, opened 2015-06-16
Solve #164, make http.py return ordered headers.

sunhao2014:
Solve #164, make http.py return ordered headers, also add the related unit test and setup dependencies.

BTW, according to Kiran's suggestion, some invalid http header test cases were added.

### DIFF
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -288,6 +288,20 @@ def test_invalid_header():
     assert r.uri == '/main/redirect/ab/1,295,,00.html'
     assert r.headers['content-type'] == 'application/x-www-form-urlencoded'
     
+    # messy header.
+    s_messy_header = 'aaaaaaaaa\r\nbbbbbbbbb'
+    try:
+        r = Request(s_messy_header)
+    except dpkt.UnpackError:
+        assert True
+    # If the http request is built successfully or raised exceptions
+    # other than UnpackError, then return a false assertion.
+    except:
+        assert False
+    else:
+        assert False
+         
+    
 def test_header_order():
     s = 'POST /main/redirect/ab/1,295,,00.html HTTP/1.0\r\n' \
     'Referer: http://www.email.com/login/snap/login.jhtml\r\n' \
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -4,11 +4,12 @@
 
 import cStringIO
 import dpkt
+import collections
 
 
 def parse_headers(f):
     """Return dict of HTTP headers parsed from a file object."""
-    d = {}
+    d = collections.OrderedDict()  # keep the order of the header.
     while 1:
         # The following logic covers two kinds of loop exit criteria.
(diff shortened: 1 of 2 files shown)


sunhao2014:
Just noticed that the PR failed the auto build.
image
The reason why py26 didn't work is we used the `collections.OrderedDict` which is supported by py27. Thus how do we deal with this problem? Should we add support to Python versions older than 2.7?

kbandla:
We could have, if ordereddict was a _necessity_. However:
* many users probably dont need the exact header order
* python2.6 is still widely used

How about doing this instead:
`try` to import collections. On `ImportError`, use a standard python dict. If not, use `collections.OrderedDict`

brifordwylie:
We could pull https://pypi.python.org/pypi/ordereddict as a dependency. One of the nice things about dpkt is we have no requirements outside the standard library, seems a shame to add a dependency just for this small thing.

brifordwylie:
Just saw Kiran's comment, yeah that's the right thing to do,

dpkt/http.py
```diff
@@ -4,11 +4,12 @@
 
 import cStringIO
 import dpkt
+import collections
```
kbandla: try:
    from collections import OrderedDict
except ImportError:
    OrderedDict = dict

dpkt/http.py
```diff
@@ -4,11 +4,12 @@
...
 
 
 def parse_headers(f):
     """Return dict of HTTP headers parsed from a file object."""
-    d = {}
+    d = collections.OrderedDict()  # keep the order of the header.
```
kbandla: d = OrderedDict()

Saylenty:
Guys, is it really usefull feature? If we use ordered dict we would lose in speed of execution (and I think even in seconds). Once, I tried to optimise one problem (https://github.com/Saylenty/algorithm_with_dots-matrices) trying to solve it using ordered set, but this was not a good idea.

obormot:
@Saylenty +1

kbandla:
Well, it is true too. OrderedDict is written in python (vs C). 

What if we do this instead? (This will not break any existing code either)
```
parse_headers(f, ordered=False):
    d = {}
    if ordered:
        from collections import OrderedDict
        d = OrderedDict()
   ....
```

obormot:
@kbandla I like this option...  but tbh my favorite solution is to not implement this at all :-)   the HTTP protocol spec does not dictate any order in the headers, so dpkt is solid there. If someone wants them in order for their specific use case then it's easy to write their own parse function and feed it tcp.data.

kbandla:
:+1: @obormot lets not implement this. go ahead and remove it @sunhao2014 (ordereddict support)

sunhao2014:
Sure. I'll close this PR and send a new PR that only contains the added http invalid test cases.

Saylenty:
@kbandla , Personally, i would rather implement this feature using OrderedDIct as you've written and write a warning only in the documentation that it can be much slower...

PR #191, opened 2015-06-17
Add invalid http header test case.

sunhao2014:
A reduction of #188, only retains the invalid http header test case.


kbandla:
:+1:

brifordwylie:
looks good, merging...

PR #192, opened 2015-06-17
enhance Packet.__repr__ to display dynamically added fields;  minor pep8 fixes

obormot:
Before the change repr() on any decoded packet would only show fields defined in the ```__hdr__```, and none of fields dynamically added in unpack(). Example:
```
In [1]: import vrrp
In [2]: vrrp.VRRP('\x21\x01\x64\x01\x00\x01\xba\x52\xc0\xa8\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00')
```
before:
```
Out[2]: VRRP(count=1, advtime=1, sum=47698, vrid=1, priority=100)
```
after:
```
Out[2]: VRRP(count=1, advtime=1, sum=47698, vrid=1, priority=100, addrs=['\xc0\xa8\x00\x01'], auth='\x00\x00\x00\x00\x00\x00\x00\x00')
```

### DIFF
--- a/dpkt/dpkt.py
+++ b/dpkt/dpkt.py
@@ -9,16 +9,20 @@ import struct
 import array
 
 
-class Error(Exception): pass
+class Error(Exception):
+    pass
 
 
-class UnpackError(Error): pass
+class UnpackError(Error):
+    pass
 
 
-class NeedData(UnpackError): pass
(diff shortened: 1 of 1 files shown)


obormot:
Actually I'd like to work a bit more on this, pulling back for now

PR #201, opened 2015-06-30
Fix #77: HTTP responses with no body cause other responses to be consumed.

### LINKED ISSUE #77: HTTP responses with no body cause other responses to be consumed
_From [andrewf...@gmail.com](https://code.google.com/u/110748771224948697237/) on September 27, 2010 18:40:27_

Put more clearly, when an HTTP response, say, 304 Not Modified, has no body but still has a content-type header, all data after that in the stream is consumed. What steps will reproduce the problem? 1. Unpack attached zip file
2. Run dpkt_bug.py, which attempts to construct dpkt.http.Response's with the data in the file stream.txt, included.

This program prints the number of responses parsed. There are two responses in the file, but only one is detected, with the other response as its body. You can see this if you print the responses instead of just the length of the list.

This test was run on Windows Vista with dpkt 1.7.

**Attachment:** [dpkt_http_bug.zip](http://code.google.com/p/dpkt/issues/detail?id=50)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=50_

sunhao2014:
Please see detailed problem description and discusstions on #77.

### DIFF
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -182,7 +182,7 @@ class Response(Message):
         # (informational), 204 (no content), and 304 (not modified) responses
         # MUST NOT include a message-body. All other responses do include a
         # message-body, although it MAY be of zero length.
-        is_body_allowed = int(self.status) >= 200 and 204 != int(status) != 304
+        is_body_allowed = int(self.status) >= 200 and 204 != int(self.status) != 304
         Message.unpack(self, f.read(), is_body_allowed)
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


kbandla:
Nice suggestions all. I am glad we have eyes watching for the smallest performance hit!

About Iandscape: short answer is that - it is okay to have multiple lines.  `max-line-length` the maximum allowed line length after which landscape will complain. But the guiding principle is to have as many lines as necessary to make the code readable. Having complex logic in one line will, most times, make it hard to read. In some cases, there is a speed advantage of having long lines (eg list comprehensions, etc). But most times, there is no speedup. This is the case with most projects, not just with dpkt. 

About the additional `int` : it is probably not going to hurt as much on a L7 module - as it gets called much less often than a L3 module's code. However, we might as well fix this, as we know about it.

sunhao2014:
Many thanks to you all. We've had a nice discussion on the code line length and "`int` issue". I've updated the code and pushed a new commit according to saylenty's suggestion.

Issue #141, opened 2015-06-04
import IP protocols can import modules from other packages and fail.


kbandla:
_From [de...@chromium.org](https://code.google.com/u/117933008161009266207/) on October 10, 2013 14:30:25_

If you have installed a python module somewhere on your path that has the same name as a protocol listed on the ip.py module (like IP_PROTO_ROUTING attempts to import the "routing" module) your module could be imported in dpkt instead of the intended one, and latter fail because protocol.PROTOCOL is expected.

To fix this, we need to pass levels=0 to the __import__ on ip.py. Attached is the patch to fix this with an extra check.

Thanks :)

**Attachment:** [ip-import-fix.patch](http://code.google.com/p/dpkt/issues/detail?id=115)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=115_

kbandla:
_From [de...@chromium.org](https://code.google.com/u/117933008161009266207/) on October 21, 2013 19:16:22_

Hi again!
The actual fix for this problem is passing level=1 to the __import__ .

See the patch we added in chromium: https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/HEAD/dev-python/dpkt/files/dpkt-1.8-fix-ip-import.patch

bitonio:
Same issue happens here, even with level=1:

```
2015-06-20 00:32:22,619 [ERROR] [113:start:MainThread] 'module' object has no attribute 'AN'
Traceback (most recent call last):
[...]
    import dpkt
  File "/usr/local/lib/python2.7/dist-packages/dpkt/__init__.py", line 23, in <module>
    import ethernet
  File "/usr/local/lib/python2.7/dist-packages/dpkt/ethernet.py", line 134, in <module>
    __load_types()
  File "/usr/local/lib/python2.7/dist-packages/dpkt/ethernet.py", line 127, in __load_types
    mod = __import__(modname, g, level=1)
  File "/usr/local/lib/python2.7/dist-packages/dpkt/ip6.py", line 135, in <module>
    import ip
  File "/usr/local/lib/python2.7/dist-packages/dpkt/ip.py", line 309, in <module>
    __load_protos()
  File "/usr/local/lib/python2.7/dist-packages/dpkt/ip.py", line 305, in __load_protos
    IP.set_proto(v, getattr(mod, name.upper()))
AttributeError: 'module' object has no attribute 'AN'
```

kbandla:
[...]

Issue #61, opened 2015-06-04
dynamic module loading too careless


kbandla:
_From [federico...@gmail.com](https://code.google.com/u/109927747085953866358/) on June 20, 2010 10:02:19_

If pigment python is installed, dpkt 1.7 fails with the error below:

Traceback (most recent call last):
  File "./setup.py", line 4, in \<module>
    import dpkt
  File "/usr/obj/i386/py-dpkt-1.7/dpkt-1.7/dpkt/__init__.py", line 24, in \<module>
    import ethernet
  File "/usr/obj/i386/py-dpkt-1.7/dpkt-1.7/dpkt/ethernet.py", line 128, in \<module>
    __load_types()
  File "/usr/obj/i386/py-dpkt-1.7/dpkt-1.7/dpkt/ethernet.py", line 122, in __load_types
    mod = __import__(modname, g)
  File "/usr/obj/i386/py-dpkt-1.7/dpkt-1.7/dpkt/ip6.py", line 95, in \<module>
    import ip
  File "/usr/obj/i386/py-dpkt-1.7/dpkt-1.7/dpkt/ip.py", line 255, in \<module>
    __load_protos()
  File "/usr/obj/i386/py-dpkt-1.7/dpkt-1.7/dpkt/ip.py", line 252, in __load_protos
    IP.set_proto(v, getattr(mod, name.upper()))
AttributeError: 'module' object has no attribute 'PGM'

This was found in OpenBSD 4.7/x86 but the error is not OS specific.
Attached diff should fix it.
Thanks.

**Attachment:** [dpkt.diff](http://code.google.com/p/dpkt/issues/detail?id=34)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=34_

kbandla:
_From [e.najda...@gmail.com](https://code.google.com/u/103405961500002997098/) on May 12, 2012 07:55:55_

Suggested solution is not good, because it will just mask the problem.  Better to add level=1 parameter to all imports (see diff).

**Attachment:** [dpkt-imports.diff](http://code.google.com/p/dpkt/issues/detail?id=34#hc2)

kbandla:
_From [ki...@kiddaland.net](https://code.google.com/u/111743600165148955252/) on December 16, 2013 13:22:23_

changing the line to 
                mod = __import__('dpkt.{}'.format(name), g)
should also fix it...

kbandla:
[...]

PR #203, opened 2015-07-04
Fix dynamic imports

kbandla:
This is a fix for #61 and #141. We now make sure that modules-to-import are searched 'relative to the directory of the module'.

### DIFF
--- a/dpkt/ethernet.py
+++ b/dpkt/ethernet.py
@@ -124,10 +124,10 @@ def __load_types():
             name = k[9:]
             modname = name.lower()
             try:
-                mod = __import__(modname, g)
-            except ImportError:
+                mod = __import__(modname, g, level=0)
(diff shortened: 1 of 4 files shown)


dpkt/ethernet.py
```diff
@@ -124,10 +124,10 @@ def __load_types():
             name = k[9:]
             modname = name.lower()
             try:
-                mod = __import__(modname, g)
-            except ImportError:
+                mod = __import__(modname, g, level=0)
```
sunhao2014: I wonder why in `ethernet.py` the modification is `level=0` while other modules are all `level=1`.

kbandla: good catch. Fix coming up shortly

kbandla:
please review and merge if it looks good

kbandla:
Looks like its working, based on #141. Merging.

PR #213, opened 2015-08-15
Fix #199: ImportError: No module named test.

### LINKED ISSUE #199: ImportError: No module named test
Some people using the dpkt-master have experienced this issue:
```
>>> import test
  File "<stdin>", line 1, in <module>
ImportError: No module named test
```
This could occur because of many different reasons. We need to have a fallback when we fail to find pystone.

sunhao2014:
Since we never use the `duration` decorator in the project. Just simply remove the `duration` decorator and related test to solve the issue.

### DIFF
--- a/dpkt/decorators.py
+++ b/dpkt/decorators.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import warnings
 from timeit import Timer
-from test import pystone
 from time import sleep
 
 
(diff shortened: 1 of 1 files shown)


kbandla:
Yeah, @saylenty was testing decorator performance early on. We can safely merge this pr

saylenty:
If we or smb don't use this decorator, I think pr should be marged. :+1:

kbandla:
You can also get rid of the other two import statements (timeit and sleep). Then merge.

sunhao2014:
Sure, the two imports has been removed. Merging.

Issue #146, opened 2015-06-04
[patch] dns.unpack_name improvements


kbandla:
_From [george.m...@gmail.com](https://code.google.com/u/114956796509061408799/) on January 15, 2014 09:41:10_

In my research, I ran across a sample that sends random data over UDP port 53, and dpkt.dns.DNS was raising IndexError when this happened.  I sat down and improved the unpack_name implementation.

The new implementation:
    * is not recursive.
    * detects pointer loops as soon as possible and raises UnpackError.
    * always raises dpkt.UnpackError when the packet is malformed in some way.
    * uses '.'.join() to build the return value instead of repeatedly appending to a string.

This resulted in a series of patches (unit tests and the actual implementation).

**Attachment:** [0001-Unit-test-demonstrating-IndexError-for-DNS-random-da.patch 0002-Add-test-case-for-circular-pointers.patch 0003-Add-unit-test-for-very-long-names.patch 0004-Improved-unpack_name-implementation.patch](http://code.google.com/p/dpkt/issues/detail?id=120)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=120_

brifordwylie:
@sunhao2014 so does PR #214 close this issue, if so please close.

PR #214, opened 2015-08-17
Add dns.unpack_name improvements and related tests.

sunhao2014:
This update is totally based on the patch of #146. There are three improvements for `dns.py`.

1. Changes the `unpack_name` from recursive to iterative when there are `ptr`s in the `name` field.
1. Detects pointer loops as soon as possible and raises `UnpackError`.
1. Uses `'.'.join()` to build the return value instead of repeatedly appending to a string to improve the performance.

I checked the logic and run the updated code, it seems good. Also it could pass all tests in `dns.py` module.

### DIFF
--- a/dpkt/dns.py
+++ b/dpkt/dns.py
@@ -88,7 +88,7 @@ def unpack_name(buf, off):
     saved_off = 0
     start_off = off
     while True:
-        if off > len(buf):
+        if off >= len(buf):
             raise dpkt.NeedData()
         n = ord(buf[off])
         if n == 0:
--- a/dpkt/dns.py
+++ b/dpkt/dns.py
@@ -83,28 +83,37 @@ def pack_name(name, off, label_ptrs):
 
 
 def unpack_name(buf, off):
-    name = ''
+    name = []
+    name_length = 0
     saved_off = 0
-    for _ in range(100):  # XXX
+    start_off = off
+    while True:
+        if off > len(buf):
+            raise dpkt.NeedData()
         n = ord(buf[off])
         if n == 0:
             off += 1
             break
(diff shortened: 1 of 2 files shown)


dpkt/dns.py
```diff
@@ -83,28 +83,37 @@ def pack_name(name, off, label_ptrs):
...
 
 def unpack_name(buf, off):
-    name = ''
+    name = []
+    name_length = 0
     saved_off = 0
-    for _ in range(100):  # XXX
+    start_off = off
+    while True:
+        if off > len(buf):
```
gmacon: I've just noticed that this should be `if off >= len(buf)` since we index buf by off on line 93.  Sorry about that.

brifordwylie: @sunhao2014 can you look at/address/comment on @gmacon comment above? After that please go ahead and merge.

sunhao2014: Yes, should be `off >= len(buf)`. Updated and merging. Thanks, @gmacon and @brifordwylie.

sunhao2014:
@brifordwylie, the `continuous-integration/travis-ci/push` build failed, however I'm not sure why. Related log is as follows.
`You have to provide either repo_token in .coveralls.yml, or launch via Travis or CircleCI
ERROR: InvocationError: '/home/travis/build/kbandla/dpkt/.tox/coveralls/bin/coveralls'`
Could you please take a look at this issue?

brifordwylie:
Just did a cherry-pick dd4ec61 for tests/coveralls to pass... (newer tox version requires this env var for Travis).

sunhao2014:
Aha, got it! The branch has been deleted.

Issue #86, opened 2015-06-04
Payload protocol checksum calculation broken for IPv6 parsing


kbandla:
_From [aa...@monkey.org](https://code.google.com/u/116105672249653840815/) on January 07, 2011 09:41:28_

There are a couple problems.  One is that you don't want to look at ip6.nxt to check the layer 4 protocol, because if there are extension headers, you need to walk the chain to find it.  That is already done in unpack(), where ip6.p is set to the payload protocol, so we can use that.

Perhaps this is a separate issue, but there is also a bug in the icmp6 unpack, which causes checksum calculation failure for icmp6 in many cases.  The problem is dpkt doesn't have handlers for all possible icmp6 types (for example, missing important ones like neighbour solitication == 135).  In this case the key lookup fails in _typesw, but you don't want to set self.data, because this will throw off the length calculation and you will end up with bad icmp6 checksums.

Note: The second problem was fixed for v4 over 3 years ago in icmp.py by jono (via r45 ).

Diff attached w/ test.

**Attachment:** [dpkt-ipv6-cksum-fill.diff](http://code.google.com/p/dpkt/issues/detail?id=59)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=59_

sunhao2014:
Second problem fixed on 1e9600f. First problem still remains.

PR #215, opened 2015-08-17
Fix #104: icmp6 type/code slip duplicate.

### LINKED ISSUE #104: icmp6 type/code slip duplicate?
_From [ruy.su...@gmail.com](https://code.google.com/u/102789873700722722337/) on September 27, 2011 10:32:33_

What steps will reproduce the problem? >>> print binascii.b2a_hex(buf)
002314008b3c000c29ea6fad86dd6000000000203afffe80000000000000020c29fffeea6fadfe80000000000000655dc7efb798391387000be800000000fe80000000000000655dc7efb79839130101000c29ea6fad
\>>> frame = dpkt.ethernet.Ethernet(buf)
\>>> print binascii.b2a_hex(str(frame))
002314008b3c000c29ea6fad86dd6000000000203afffe80000000000000020c29fffeea6fadfe80000000000000655dc7efb798391387000be887000be800000000fe80000000000000655dc7efb79839130101000c29ea6fad        

icmp6 type and code and the back 4bytes is duplicate unpacked. What is the expected output? What do you see instead? The above. What version of the product are you using? On what operating system? First line of icmp6.py
\# $Id: icmp6.py 23 2006-11-08 15:45:33Z dugsong $ Please provide any additional information below. ***** C:\DEV\PYTHON\ICMP6_ORIGINAL.PY
   71:          except (KeyError, dpkt.UnpackError):
   72:              self.data = buf
***** C:\DEV\PYTHON\ICMP6.PY
   71:          except (KeyError, dpkt.UnpackError):
   72:              pass
*****

Same icmp.py.
Maybe work.

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=77_

sunhao2014:
See more details about this issue on #104.
The reason why we have this problem is mentioned in #86.

>The problem is dpkt doesn't have handlers for all possible icmp6 types (for example, missing important ones like neighbour solitication == 135). In this case the key lookup fails in _typesw, but you don't want to set self.data, because this will throw off the length calculation and you will end up with bad icmp6 checksums.

Actually we have fixed the similar problem in `icmp.py` many years ago (see http://code.google.com/p/dpkt/source/detail?r=45). But in `icmp6.py` the problem is still remaining. This PR simply fix this issue in `icmp6.py`.

By trying the test script below, we could verify the validity of this update.

``` python
from binascii import *
from dpkt import ethernet

buf = "\x00\x23\x14\x00\x8b\x3c\x00\x0c\x29\xea\x6f\xad\x86\xdd\x60\x00\x00\x00\x00\x20\x3a\xff\xfe\x80\x00\x00\x00\x00\x00\x00\x02\x0c\x29\xff\xfe\xea\x6f\xad\xfe\x80\x00\x00\x00\x00\x00\x00\x65\x5d\xc7\xef\xb7\x98\x39\x13\x87\x00\x0b\xe8\x00\x00\x00\x00\xfe\x80\x00\x00\x00\x00\x00\x00\x65\x5d\xc7\xef\xb7\x98\x39\x13\x01\x01\x00\x0c\x29\xea\x6f\xad"
print b2a_hex(buf)
frame = ethernet.Ethernet(buf)
print b2a_hex(str(frame))
```

### DIFF
--- a/dpkt/icmp6.py
+++ b/dpkt/icmp6.py
@@ -74,4 +74,4 @@ class ICMP6(dpkt.Packet):
             self.data = self._typesw[self.type](self.data)
             setattr(self, self.data.__class__.__name__.lower(), self.data)
         except (KeyError, dpkt.UnpackError):
-            self.data = buf
+            pass


kbandla:
Looks good to me. Can be merged.

PR #216, opened 2015-08-19
Fix #91: Fragment offset values for IP packets incorrect.

### LINKED ISSUE #91: Fragment offset values for IP packets incorrect
_From [j...@midokura.jp](https://code.google.com/u/118443755321160583317/) on February 03, 2011 01:44:51_

What steps will reproduce the problem? 1. Run attached sample program, which parses an IP packet with fragment offset 0. 2. 3. What is the expected output? What do you see instead? Expect .off of the dpkt parsed IP packet to be 0.
Actual .off of the dpkt parsed IP packet is 16384. What version of the product are you using? On what operating system? 1.6 (1.6+svn54-1) on Ubuntu 10.10 (Maverick) Please provide any additional information below.

**Attachment:** [ipoff.py](http://code.google.com/p/dpkt/issues/detail?id=64)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=64_

sunhao2014:
Please check out #91 to have a look at the original issue.

The reason why we have this problem is in `ip.py`, there is a composite attribute `off`, it contains 4 IP fragmentation related attributes, respectively `rf`, `df`, `mf` and `offset`. Just like the strategy we deal with `_v_hl`, I added setters and getters to those attributes together with the test case.

### DIFF
--- a/dpkt/ip.py
+++ b/dpkt/ip.py
@@ -45,6 +45,38 @@ class IP(dpkt.Packet):
     @hl.setter
     def hl(self, hl):
         self._v_hl = (self._v_hl & 0xf0) | hl
+        
+    @property
+    def rf(self):
+        return (self.off >> 15) & 0x1
+
+    @rf.setter
+    def rf(self, rf):
+        self.off = (self.off & ~IP_RF) | (rf << 15)
+        
+    @property
+    def df(self):
+        return (self.off >> 14) & 0x1
+
+    @df.setter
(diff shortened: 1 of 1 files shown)


brifordwylie:
@sunhao2014 looks good, please go ahead and feel free to merge these PR's yourself.

sunhao2014:
Merging, thanks for reviewing, Brian.

Issue #229, opened 2015-12-09
pcap.py: DLT_RAW constant value


WGH-:
Why DLT_RAW is defined to be 14 or 12? In my experience of running tcpdump on tun-interface, it produced pcap files containing just IPv4 packets (i.e. no Ethernet or other kind of encapsulation) with datalink type of 101 (0x65), which was consistent with tcpdump.org definition of DLT_RAW (http://www.tcpdump.org/linktypes.html)

RemiDesgrange:
Maybe need to close this ?

WGH-:
Uhm, no, it's not really fixed.

https://github.com/kbandla/dpkt/blob/c1450e2a6c81a9ff8e82776f91b2b1763cd814bf/dpkt/pcap.py#L128 - these lines are still there, and they override the definitions above.

grumple-higi:
Yeah I just ran into this today, curious why this was closed as DLT_RAW is defined as 101 which is what I have found it to be in almost all files, yet it get's overridden to either 12 or 14 after all the definitions.  So I have to compare against 101 to get the functionality working vs using a constant.

PR #250, opened 2016-03-08
Add some DLT in pcap.py

RemiDesgrange:
I added some DLT const in  pcap.py, maybe we can copy them to pcapng too, but i do not like this option.

And this is closing the #229


RemiDesgrange:
wooops sorry, I modified print_packet to get it working on widnows, but need another PR...

kbandla:
Looks good to me. Unless anyone has any comments, we can merge this.

PR #251, opened 2016-03-10
nanosecond format improvements

ecatmur:
First, using Python float (IEEE 64-bit double) has insufficient precision for representing nanosecond timestamps; right now float has a precision of 2**-22 s = 238.4 ns, which is much coarser than the precision of high-resolution timestamping equipment. Suggest using decimal.Decimal; numpy.datetime64 would also work but probably not worth the dependency.

Second, support writing in nanosecond format as well as reading.

### DIFF
--- a/dpkt/pcap.py
+++ b/dpkt/pcap.py
@@ -5,6 +5,7 @@
 import sys
 import time
 import dpkt
+from decimal import Decimal
 
 
 TCPDUMP_MAGIC = 0xa1b2c3d4L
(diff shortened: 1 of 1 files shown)


obormot:
looks good!

kbandla:
Testing patch..

brifordwylie:
@kbandla lets go ahead and merge this if we can.. it's been 'hanging in the wind' for a while...

Issue #148, opened 2015-06-04
FCS field on dpkt.ethernet.Ethernet


kbandla:
_From [oscar....@gmail.com](https://code.google.com/u/117531032706262214992/) on March 12, 2014 11:30:07_

What steps will reproduce the problem? 1. Suppose you have a string with a raw Ethernet packet, *including* the FCS bytes on the last 4 positions.
2. p = dpkt.ethernet.Ethernet(raw_ethernet_packet_string)
3. p.data has all the packet payload but FCS bytes are lost/not saved. What is the expected output? What do you see instead? I expect to have FCS data somewhere on Ethernet packet object. There isn't any FCS field to look. What version of the product are you using? On what operating system? dpkt 1.6 on debian jessie Please provide any additional information below. I'm not sure if this is a needed feature to other users (in fact, most of the HW and pcap files don't return FCS field to software layers), my need for FCS support is somehow special (I'm programming a MAC HW). However, it would be nice to have an optional FCS support. Thanks.

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=122_

kbandla:
_From [timur.al...@gmail.com](https://code.google.com/u/117790548747779378711/) on July 02, 2014 10:17:55_

Did not mean to comment on this issue

**Status:** New  
**Owner:** ---

kbandla:
_From [oscar....@gmail.com](https://code.google.com/u/117531032706262214992/) on July 08, 2014 11:41:55_

Here I put a extended class I use in my project. Feel free to use it to update Ethernet class.

Best regards.

\### Begin code ###

import dpkt
from zlib import crc32

class EthWithFcs(dpkt.ethernet.Ethernet):
    def __len__(self):
        if hasattr(self, "fcs"):
            return dpkt.ethernet.Ethernet.__len__(self) + 4
        else:
            return dpkt.ethernet.Ethernet.__len__(self)
            
    def __str__(self):
        if hasattr(self, "fcs"):
            return self.pack_hdr() + str(self.data) + self._fcs_tostr()
        else:
            return self.pack_hdr() + str(self.data)
        
    def add_fcs(self):
        self.fcs = self._fcs_calc()
        
    def is_fcs_valid(self):
        return self.fcs == self._fcs_calc()
            
    def unpack(self, buf):
        dpkt.ethernet.Ethernet.unpack(self, buf)
        if dpkt.ethernet.Ethernet.__len__(self) \< len(buf):
[...]

Issue #140, opened 2015-06-04
Unable to decode packets with both VLAN tag and MPLS label


kbandla:
_From [marzetti...@gmail.com](https://code.google.com/u/110232804116673833794/) on October 01, 2013 03:37:05_

Even with last CVS dpkt is unable to decode packets that carry both VLAN tag and MPLS label.

I attached a "patch" to fix the issue.

**Attachment:** [ethernet.patch](http://code.google.com/p/dpkt/issues/detail?id=114)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=114_

obormot:
cleanup.. this was addressed, closing

Issue #89, opened 2015-06-04
problem with mpls decapsulation


kbandla:
_From [toby.bea...@gmail.com](https://code.google.com/u/104115839347211898201/) on January 23, 2011 17:06:35_

What steps will reproduce the problem? 1. Analyzing mpls encapsulated packets What is the expected output? What do you see instead? I open a pcap file with multiple mpls layers on each packet.  When I try to iterate through, it has eth.data as a type str instead of dpkt.ip.IP.  If I try forcing ip on eth(data) I get "dpkt.dpkt.UnpackError: invalid header length"

for ts, data in pcapreader:
    try: eth = dpkt.ethernet.Ethernet(data)
    except: continue
    print eth.type
    ip = dpkt.ip.IP(eth.data)

2048
Traceback (most recent call last):
  File "pcap-parse.py"
, line 14, in <module>
    ip = dpkt.ip.IP(eth.data)
  File "c:\Python27\lib\site-packages\dpkt\dpkt.py", line 72, in **init**
    self.unpack(args[0])
  File "c:\Python27\lib\site-packages\dpkt\ip.py", line 56, in unpack
    raise dpkt.UnpackError, 'invalid header length'
dpkt.dpkt.UnpackError: invalid header length What version of the product are you using? On what operating system? 1.7 Please provide any additional information below.

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=62_

kbandla:
_From [toby.bea...@gmail.com](https://code.google.com/u/104115839347211898201/) on January 24, 2011 14:25:23_

Seems to be breaking in ip.py during the unpack.

def unpack(self, buf):
        dpkt.Packet.unpack(self, buf)
        ol = ((self.v_hl & 0xf) \<< 2) - self.__hdr_len__
        if ol \< 0:  

in my debug self.v_hl = 1, self.__hdr_len__ = 20, and ol = -16.

obormot:
cleaning up old issues
this is likely fixed with multiple MPLS improvements made and unit tests added

Issue #83, opened 2015-06-04
Packet from dpkt.ethernet.Ethernet constructor is unparsed.


kbandla:
_From [googa...@jlm.ofb.net](https://code.google.com/u/110220263730420990331/) on December 09, 2010 05:16:20_

What steps will reproduce the problem? ### Begin sample code ###
from dpkt import ethernet
from dpkt import arp

arp_pkt = arp.ARP(sha='123456', spa='abcd', tha='7890ab', tpa='wxyz')
eth_pkt = ethernet.Ethernet(data=str(arp_pkt), type=ethernet.ETH_TYPE_ARP,
                src='ABCDEF', dst='PQRSTU')
print "Is .arp in eth_pkt? &#37;s" &#37; ('arp' in eth_pkt.__dict__)
eth_pkt.unpack(str(eth_pkt))
print "Is .arp in eth_pkt? &#37;s" &#37; ('arp' in eth_pkt.__dict__)
\### End sample code ### What is the expected output? What do you see instead? Expect .arp in the constructed Ethernet packet, without a need to call
.unpack on its own str().  Actual .arp is not in constructed Ethernet packet, but .unpack on its own str() will generate it. What version of the product are you using? On what operating system? dpkt 1.6 on Ubuntu 10.10 (package 1.6+svn54-1) Please provide any additional information below.

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=56_

PR #267, opened 2016-05-01
Ethernet and LLC improvements

### LINKED ISSUE #59: VLAN tagging of Ethernet frames
_From [mattjgal...@gmail.com](https://code.google.com/u/103561754316806446990/) on March 18, 2010 06:06:35_

It would be great if there was support for VLAN tagging of Ethernet frames.
There seems to be support for understanding a tagged packet, but not for
actually tagging them.

I've roughly added support myself, but no where near ready for a patch. But
I will keep going and hopefully submit something.

I just wanted to lodge here in case it was already being worked on?

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=32_

obormot:
Multiple fixes and enhancements to Ethernet and LLC decoders. Summary of changes:
+ VLAN tags are organized into 2 new classes: VLANtag8021Q and VLANtagISL, contained in a new list field "Ethernet.vlan_tags";
+ MPLS labels are similarly a new class MPLSlabel, contained in a new list field "Ethernet.mpls_labels";
+ support IEEE 802.1Q double tagging (aka QinQ);
+ Ethernet/IEEE 802.3 Ethernet can be constructed with MPLS labels and/or VLAN tags;
+ fixed construction of LLC with SAP/SNAP headers and data encapsulation;
+ fixed parsing of Cisco ISL VLAN tag (was off by 1 bit);
+ correctly compute len() of Ethernet with tags/labels and LLC;
+ compute 802.3 Ethernet Frame Check Sequence (FCS), recognize trailer bytes;
+ added a bunch of unit tests.

Changes are backward compatible, except the following:
+ LLC is now a layer. This matches Wireshark representation, and I believe it was meant to be this way originally in dpkt, but eventually llc.py unpacking code was moved into ethernet.py. I guess it was moved to deal with circular referencing between the two modules. I found a solution that avoids these issues.
+ As a result of the above change, LLC.data that is now properly decoded to the next layer (previously one had to explicitly cast it, e.g. ip.IP(llc.data)).

Fixes #59, #83, #89, #140, #148


brifordwylie:
Looks like excellent work as always. Given the detailed nature of this work I'm not qualified to review it.. @kbandla do you have some cycles? If so great, otherwisel @obormot can merge as his discretion.

kbandla:
Nice PR. Based on a quick look, everything looks good. I will go over the details this weekend

PR #274, opened 2016-06-27
change variable name in rip __len__ function

j-m-patterson:
the __len__ function in rip.py used a variable named 'len' which, as a
buit-in function, caused problems.


obormot:
looks good

PR #280, opened 2016-08-12
Minor fixes in various protocols.

kleptog:
- Add IPIP (which is the same as IP)
- Endianness in GRE fields
- Correct PPP encapsulation header

Add tests.

Based on decoding http://packetlife.net/captures/gre_and_4over6.cap

### DIFF
--- a/dpkt/gre.py
+++ b/dpkt/gre.py
@@ -97,7 +97,7 @@ class GRE(dpkt.Packet):
         if fields:
             fmt = ''.join(fmts)
             fmtlen = struct.calcsize(fmt)
-            vals = struct.unpack(fmt, self.data[:fmtlen])
+            vals = struct.unpack("!" + fmt, self.data[:fmtlen])
             self.data = self.data[fmtlen:]
             self.__dict__.update(dict(zip(fields, vals)))
         if self.flags & GRE_RP:
@@ -109,8 +109,12 @@ class GRE(dpkt.Packet):
                 if not sre.len:
                     break
             self.sre = l
-        self.data = ethernet.Ethernet._typesw[self.p](self.data)
-        setattr(self, self.data.__class__.__name__.lower(), self.data)
+        try:
+            self.data = ethernet.Ethernet._typesw[self.p](self.data)
+            setattr(self, self.data.__class__.__name__.lower(), self.data)
+        except (KeyError, dpkt.UnpackError):
+            # data alrady set
+            pass
 
     def __len__(self):
         opt_fmtlen = struct.calcsize(''.join(self.opt_fields_fmts()[1]))
@@ -131,3 +135,31 @@ class GRE(dpkt.Packet):
 import ethernet
 
 GRE._protosw.update(ethernet.Ethernet._typesw)
+
+
+def test_gre_v1():
+    # Runs all the test associated with this class/file
+    s = "3081880a0067178000068fb100083a76".decode('hex') + "A" * 103
+    g = GRE(s)
+
+    assert g.v == 1
+    assert g.p == 0x880a
+    assert g.seq == 430001
+    assert g.ack == 539254
+    assert g.callid == 6016
+    assert g.len == 103
+    assert g.data == "A" * 103
+
+    s = "3001880a00b2001100083ab8".decode('hex') + "A" * 178
+    g = GRE(s)
+
+    assert g.v == 1
+    assert g.p == 0x880a
+    assert g.seq == 539320
+    assert g.callid == 17
+    assert g.len == 178
+    assert g.data == "A" * 178
+
+
+if __name__ == '__main__':
+    test_gre_v1()
--- a/dpkt/ipip.py
+++ b/dpkt/ipip.py
@@ -1,1 +1,3 @@
+# Defines a copy of the IP protocol as IPIP so the protocol parsing in ip.py
+# can decode IPIP packets.
 from ip import IP as IPIP
--- a/dpkt/ipip.py
+++ b/dpkt/ipip.py
@@ -1,0 +1,1 @@
+from ip import IP as IPIP
--- a/dpkt/ppp.py
+++ b/dpkt/ppp.py
@@ -26,6 +26,8 @@ class PPP(dpkt.Packet):
     """
     
     __hdr__ = (
+        ('addr', 'B', 0xff),
+        ('cntrl', 'B', 3),
         ('p', 'B', PPP_IP),
     )
     _protosw = {}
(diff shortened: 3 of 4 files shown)


kleptog:
Don't see why the coverage failed, will look again later.

brifordwylie:
sorry about the coverage issue... something changed somewhere.. anyway I've fixed it with this change.. https://github.com/kbandla/dpkt/pull/281/files   do you mind cherry-picking that commit?  (0a16746a2e)

kleptog:
That fixed it, thanks!

kleptog:
This patch depends on the previous one, will rebase when the other is merged.

dpkt/gre.py
```diff
@@ -88,7 +88,7 @@ def unpack(self, buf):
         if fields:
             fmt = ''.join(fmts)
             fmtlen = struct.calcsize(fmt)
-            vals = struct.unpack(fmt, self.data[:fmtlen])
+            vals = struct.unpack("!" + fmt, self.data[:fmtlen])
```
brifordwylie: So the addition of the "!" is making sure the format is treated as network order, super weird that wasn't there before (I guess that's why you're fixing it :)

kleptog: That's deliberate. Besides that in general network protocols use network order, this makes the output match that of Wireshark.

dpkt/ipip.py
```diff
@@ -0,0 +1 @@
+from ip import IP as IPIP
```
brifordwylie: What's the point of this? Does this fix something or just provide a shorthand way of referencing ip.IP ?

kleptog: At the beginning of ip.py there is a function __load_protos which goes through the IP_PROTO_* list and loads all the modules. Adding this file means that dpkt can decode IPIP packets. That is, IPv4 or IPv6 packets where the next header is 4 (IPIP).

I agree it's a bit of a hack, but then, so is the dynamic importing. The alternative would be to add something like:

`IP.set_proto(IP_PROTO_IPIP, IP._protosw[IP_PROTO_IP])`

somewhere, but that feels even worse.

brifordwylie: Ah, I see. Thanks for the explanation... okay sounds good to me. @kbandla or @obormot any reservations to merging this PR?  If we don't hear back will merge this tomorrow.

dpkt/ppp.py
```diff
@@ -32,7 +34,7 @@ def get_p(cls, p):
     def unpack(self, buf):
         dpkt.Packet.unpack(self, buf)
         if self.p & PFC_BIT == 0:
-            self.p = struct.unpack('>H', buf[:2])[0]
+            self.p = struct.unpack('>H', buf[2:4])[0]
```
obormot: if the buf is only 3 bytes you'll get an exception here

dpkt/pppoe.py
```diff
@@ -58,7 +58,9 @@ def unpack(self, buf):
         dpkt.Packet.unpack(self, buf)
         try:
             if self.code == 0:
-                self.data = self.ppp = ppp.PPP(self.data)
+                # We need to use the light PPP header here, because PPPoE
+                # doesn't do the normal encapsulation.
+                self.data = self.ppp = ppp.PPP_light(self.data)
```
obormot: I'm concerned that this introduces a new protocol name that's not RFC'd..  maybe not what people expect when decoding a PPPoE packet.  In Wireshark it looks like PPPoE layer encapsulating PPP layer, not PPP_light. Should PPPoE possibly define its own version of PPP instead?

dpkt/ppp.py
```diff
@@ -49,6 +51,13 @@ def pack_hdr(self):
             raise dpkt.PackError(str(e))
 
 
+class PPP_light(PPP):
+    # Light version for protocols without the usual encapsulation (PPPoE)
+    __hdr__ = (
+        ('p', 'B', PPP_IP),
```
obormot: ..and this can also be 2 bytes, right?  e.g. 0xC021 -  PPP link control protocol

dpkt/ipip.py
obormot: @brifordwylie I have a couple of questions; left comments on the diff

brifordwylie: Great thanks @obormot: I knew you'd be able to take a deeper dive on this. I'll let you do the merge on the PR when all the wrinkles get ironed out..

dpkt/ppp.py
kleptog: Ok, good point.

dpkt/pppoe.py
kleptog: Yeah, this was something I wasn't sure about. I made it a subclass of the normal PPP so it wouldn't break any code, but it does seem like pppoe.PPP being different from ppp.PPP is not an unreasonable solution.

dpkt/ppp.py
kleptog: I wouldn't know about that, this header is just the original definition of the PPP header before I changed it. There were no test cases for this before. Do you have an example of what it should match? I couldn't for example figure out whether PPPoE supports protocol compression or not.

dpkt/pppoe.py
kleptog: Thinking on this some more it looks like I broke the PPPoE completely. I'm now looking at a PPPoE trace to see if I can fix it.

dpkt/ppp.py
obormot: Search for pppoe [here](https://wiki.wireshark.org/SampleCaptures), there's a sample pcap.

dpkt/pppoe.py
obormot: :)

dpkt/pppoe.py
```diff
@@ -58,8 +59,103 @@ def unpack(self, buf):
...
 
+
+class PPP(ppp.PPP):
+    # Light version for protocols without the usual encapsulation, for PPPoE
+    __hdr__ = (
+        # Usuaully two-bytes, but while protocol compression is not recommended, it is supported
+        ('p', 'B', ppp.PPP_IP),
+    )
+
+    def unpack(self, buf):
```
obormot: I don't think you need unpack() method here at all..  since you're subclassing dpkt.PPP with a different ```__hdr__``` it should all magically work.

kleptog: You do, because the protocol expansion needs to know the length of the header to work. Similarly, the pack code needs to be special because the header is a different length. You could try to merge it into the same code but it becomes quite unreadable.

brifordwylie:
@kleptog can you fix merge conflicts
@obormot if you're satisfied that all the wrinkles got addressed give a thumbs up...

brifordwylie:
@kleptog reviews are done, please fix merge conflicts and then we'll merge. Thanks for the excellent work on this :)

kleptog:
Ok done, hope I did it right...

Issue #165, opened 2015-06-04
Check if a given dpkt-layer is part of the packet


kbandla:
_From [anders.a...@gmail.com](https://code.google.com/u/117205048115784099846/) on April 20, 2015 02:58:52_

Is there a standard method for searching the data-portion of a packet for a given layer?

I have started to build a decoder for EtherNet/IP. When I sniff packets I usually need to filter for a given layer (ex. ForwardOpen etc). I did not find a method for this so I created the following two methods in dpkt.Packet:

dpkt.Packet.HasLayer( LayerClass )
dpkt.Packet.GetLayer( LayerClass )

They are basically recursive methods that check the Type of the data attribute of the packet.

The first one return True/False if the layer is located somewhere in the current packet, and the second of course return the layer in question (The first on actually run the second). 

I'm running Python 2.7.5 on windows 7 and 1.8 of dpkt.

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=139_

obormot:
I think these would look better as ```__contains__``` and ```__getitem__``` magic methods. E.g.
```python
pkt = Ethernet(buf)
if UDP in pkt:
    udp = pkt[UDP]
```

PR #282, opened 2016-08-15
Add method to match layers by type.

kleptog:
Using the new .get() method you can quickly find the IP layer in a packet.
With the .getall() method you can retrieve a list, in case there are
multiple.

It would have been nice to use the **getitem** interface, but it is already
used for another purpose.

Fixes: #165

### DIFF
--- a/dpkt/dpkt.py
+++ b/dpkt/dpkt.py
@@ -141,6 +141,24 @@ class Packet(object):
     def __str__(self):
         return self.pack_hdr() + str(self.data)
 
+    def get(self, type_):
+        """ Find first header in packet of given type """
+        if isinstance(self, type_):
+            return self
+        if isinstance(self.data, Packet):
+            return self.data.get(type_)
+        return None
+
+    def getall(self, type_, res=None):
+        """ Find all headers in packet of given type """
+        if res is None:
+            res = []
+        if isinstance(self, type_):
+            res.append(self)
+        if isinstance(self.data, Packet):
+            return self.data.getall(type_, res=res)
+        return res
+
     def pack_hdr(self):
         """Return packed header string."""
         try:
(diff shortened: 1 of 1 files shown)


kleptog:
This patch depends on the previous one, will rebase when the other is merged.

brifordwylie:
So I really like @obormot suggestion on #165 to use the __contains__ and __getitem__ methods, which would mean this kind of intuitive usage.

```
eth = Ethernet(buf)
if UDP in eth:
    udp = pkt[UDP]
```

Right now dpkt.py already overloads the __getitem__ method for getting an attribute...

```
    def __getitem__(self, k):
        try:
            return getattr(self, k)
        except AttributeError:
            raise KeyError
```
I don't actually like the the current __getitem__ functionality. dpkt has attribute based classes and that's how everyone uses it.  So retrieving the attributes with a 'dictionary' interface just muddies the waters  (BTW: all 132 dpkt tests pass fine without the __getitem__ interface). 

This is a big change, so I'll like @kbandla and @obormot weight in here, but for my $0.02 I'm going to suggest something mildly radical.
1) Remove the current __getitem__ functionality
2) Use __getitem__ and __contains__ as @obormot suggested

Also no matter what we do I don't like the 'getall' functionality, it seems superfluous and I feel like code using it will want to know/differentiate between the 'outer' IP packet and the 'inner' IP packet, so for example.. I feel like this snippet below is intuitive and natural, I'm not sure how I would code this up with a list of 2 IPs from getall()

```
if 'IP' in eth:
     ip = eth['ip']
     if 'IP' in ip:
            print('Interesting.. I have an encapsulated IP packet')
```

obormot:
@brifordwylie I agree completely, this is how scapy does it BTW. I think we want ```__getitem__``` to return the first instance of the given layer (if there are multiple); and ```__contains__``` obviously to check if the given layer is present at all.

`getall` *could* be useful for grouping fields of the same type in a list (e.g. to make a list of SSL ciphersuites, where each is an instance of a CipherSuite class), if the protocol parser didn't group them inside a wrapper field (which it should do..).  I agree it'll confuse encapsulation. I would not add it as a core library function.

brifordwylie:
Closing this PR as it's been superseded by #584.

Thanks @obormot 🥇

PR #283, opened 2016-08-16
adding the truncated flag to the properties for dns decoder

brifordwylie:
Would like a quick AOK from @obormot or @kbandla on this before I merge...


kbandla:
lgtm @brifordwylie

PR #290, opened 2016-08-23
adding an ICMP example script

brifordwylie:


kbandla:
lgtm

brifordwylie:
You can look here for the examples.. http://dpkt.readthedocs.io/en/add-icmp-example/examples.html#examples-in-dpkt-examples

note: I'm happy to change any of the examples scripts.. cause we want to make sure they are idiomatic (meaning that's exactly what dpkt users should emulate). The one thing I'm unsure of is the use of .__class__.__name.__ when determining the ICMP packet.. if that's not quite the right way to do it.. lets change it :)

brifordwylie:
merging.. happy to have people review the examples and suggest changes.. we can make any changes in another PR

PR #294, opened 2016-08-23
Remove perf-tests

### LINKED ISSUE #278: Deprecate perf tests folder
When running:
```bash
python test-perf.py
```
I get the following error message:

`   ImportError: No module named xstruct`

Github and Google does not reveal where one can easily find this Python package.  Where can one find this `xstruct` package?

kbandla:
This code depends on legacy modules to test dpkt's performance against other
packet parsing libraries. This is no longer supported. Should fix #278


brifordwylie:
@kbandla there's a known issue with landscape 'hanging' on some jobs.. so if the landscape check doesn't come back.. just merge anyway...

brifordwylie:
Merge it

PR #302, opened 2016-09-19
Add test for gzip.

gendx:


dpkt/gzip.py
```diff
@@ -70,7 +73,7 @@ class Gzip(dpkt.Packet):
     def unpack(self, buf):
         super(Gzip, self).unpack(buf)
         if self.flags & GZIP_FEXTRA:
-            n = struct.unpack('>H', self.data[:2])[0]
+            n = struct.unpack('<H', self.data[:2])[0]
```
obormot: there's still struct.pack('>H', ...) in the pack method

gendx:
It should be fine now.

PR #303, opened 2016-09-23
Add basic support for HTTP/2

gendx:
I started to implement the new HTTP/2 protocol as per RFC 7540. For now the frames can be extracted, but specialized parsers for each frame type are not implemented yet. I also tested it with some real-world traffic.


gendx:
I think that most of the HTTP/2 protocol is implemented (except for the decoding of HEADERS, but that requires to assemble several frames at a higher level, and can be done with the `hpack` library), so it's ready for review.

gendx:
Ping. Anyone has comments about my implementation?

wmnsk:
This is nice! I have two non-critical suggestions:

* in the child classes like `PaddedFrame`, it's recommended to use `super()` to call the methods in parent class instead of using the parents' name itself.
  * additionally, `unpack()` method with no customization should not be implemented in the child classes like `DataFrame`, it can be implemented by using @property.

```python
# implement property instead of overriding 'unpack()'
@property
def payload(self):
    return self.unpadded_data
```

* the package `binascii` (and `_hexdecode` method) should be written inside `TestFrame` class, as it is used only for testing.

My review is not enough because I'm NOT a collaborator of this repo, but I think this PR is worth being merged. Could anyone review?

gendx:
Thanks for your feedback!

* I used `@property` where it is relevant, as you suggested. I didn't find other occurrences where to apply it though, e.g. in `PriorityFrame` there can be an exception thrown when parsing `self.priority`, so (correct me if I'm wrong) I don't think we can delay that to the moment when the user requests the value of `priority`.

* Regarding the use of `super`, if followed the consensus in the project, there were very few occurrences using `super`. But maybe we can make a separate issue/PR to make that uniform?

* For `_hexdecode`, I took inspiration from the `ssl.py` file, but indeed we can do better. There is no need to import anything, `.decode('hex')` should be enough.

gendx:
Also, with the new Python 3 PR, I have to check that things will be compatible.

gendx:
I made this compatible with the recent Python 3 upgrade. Could anyone else review ?

gendx:
PING @kbandla @obormot @brifordwylie

brifordwylie:
Wow, sorry.. this has been here ^forever^. Since this is adding a new protocol parser there's really little chance of regression/negative impact. Although I'm NOT qualified to review the http2 decoding after sitting here many months it's probably better to push in than not.

gendx:
Yeah \o/ It's been a while indeed, I was forgetting about it myself. With HTTP/2 becoming more and more deployed, I hope it'll be useful !

PR #304, opened 2016-09-30
Add TLS ciphersuites from IANA.

gendx:
The list of TLS ciphersuites was only a stub in `ssl_ciphersuites.py`. I updated this list according to IANA (https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml) and corresponding RFCs. I also refactored a bit the module to be more consistent with the spirit of these ciphersuites, e.g. `encoding` is now split into `cipher` (like AES_128) and `mode` (like CBC).


dpkt/ssl_ciphersuites.py
```diff
@@ -50,27 +104,541 @@ def mac_size(self):
...
+    CipherSuite(0x00ff, 'NULL', '',         'NULL',     '',     'NULL',
+        'TLS_EMPTY_RENEGOTIATION_INFO'),
+    # RFC7507
+    CipherSuite(0x5600, '', '',             '', '', '',
+        'TLS_FALLBACK'),
+    CipherSuite(0xffff, '', '',             '', '', '',
+        'UNKNOWN_CIPHER'),
+
+    # RFC2246 : TLS 1.0
+    CipherSuite(0x0000, 'NULL', '',         'NULL',     '',     'NULL'),
```
obormot: pylint is going to complain about this. if you'd like to keep this type of formatting one trick I found useful was adding spaces in values, then stripping them in ```__init__```. E.g. 
instead of 

    CipherSuite(0x0001, 'RSA', '',          'NULL',     '',     'MD5'),

do

    CipherSuite(0x0001, 'RSA', '         ', 'NULL    ', '    ', 'MD5'),

then inside ```__init__```:

    def __init__(self, code, kx, auth, cipher, mode, mac, name=None, encoding=None):
          self.code = code
          self._kx = kx.strip()
          self._auth = auth.strip()
    ...

gendx: Thanks for the comment, I didn't know about this rule. Does the current CI check it?

gendx:
This should be fixed now.

kbandla:
nice PR @gendx ! lgtm

brifordwylie:
@gendx current CI check for this branch looks good: https://landscape.io/diff/461974  @kbandla please merge when you get a chance.

gendx:
I just made a last-minute fix, the function `BY_NAME` was not returning anything.

gendx:
ping

dpkt/ssl_ciphersuites.py
```diff
@@ -14,32 +14,88 @@ class CipherSuite(object):
...
+        if self._auth == '': # for RSA
+            return self._kx
+        elif self._kx == '': # for PSK
+            return self._auth
+        else:
+            return self._kx + '_' + self._auth
+
+    @property
+    def encoding(self):
+        if self._encoding == None:
```
obormot: is None

dpkt/ssl_ciphersuites.py
```diff
@@ -14,32 +14,88 @@ class CipherSuite(object):
...
+            if self.mode == '':
+                return self.cipher
+            else:
+                return self.cipher + '_' + self.mode
+        else:
+            return self._encoding
+
+    @property
+    def name(self):
+        if self._name == None:
```
obormot: is None

dpkt/ssl_ciphersuites.py
```diff
@@ -14,32 +14,88 @@ class CipherSuite(object):
...
     * mac: message authentication code algorithm
+    * name: as in 'TLS_RSA_WITH_RC4_40_MD5', can be generated by default from
+        the other parameters
+    * encoding: encoding algorithm, defaults to cipher+mode
+    Additional members:
+    * kx_auth: kx+auth algorithm, as 'KeyExchangeAlgorithm' in RFCs
     """
 
-    def __init__(self, code, name, kx, auth, encoding, mac):
+    def __init__(self, code, kx, auth, cipher, mode, mac, name=None, encoding=None):
```
obormot: Moving the arguments around breaks compatibility with any existing code. Looks like "code" determines all other parameters, so it could be the only required argument? In this case why won't we NOT change the order of the remaining arguments, but make them all optional, instead? I.e. 
```
def __init__(self, code, name=None, kx=None, auth=None, encoding=None, mac=None):
```

dpkt/ssl_ciphersuites.py
```diff
@@ -14,32 +14,88 @@ class CipherSuite(object):
 
     Members/args:
     * code: two-byte ID code, as int
-    * name: as in 'TLS_RSA_WITH_RC4_40_MD5'
     * kx: key exchange algorithm, string
     * auth: authentication algorithm, string
-    * encoding: encoding algorithm
+    * cipher: stream or block cipher algorithm, as 'AES_128'
+    * mode: mode of operation for block ciphers, as 'CBC' or 'GCM'
```
obormot: It's nice to see these examples here, suggest you update the "kx" and "auth" docstrings with examples as well.

gendx: I will add them.

dpkt/ssl_ciphersuites.py
gendx: Indeed `code` determines all the other parameters, and the purpose of this file is to define the mapping between the TLS ciphersuite code and the other parameters, via the dictionary `BY_CODE`. The `CipherSuite` class is only an internal helper class to define everything more easily, but shouldn't be instantiated from outside the file (see line 7: "These are supposed to be immutable; don't mess with them."). So changing the order of the arguments wouldn't break existing code because users shouldn't instantiate `CipherSuite` themselves in the first place (and I cannot see how it would be useful).

Also, the previous code was only a stub and the order of the arguments was somewhat arbitrary and didn't really correspond to the hundreds of cipher suites that exist. So I decided to change the order to something more logical (e.g. `name` is really determined by the other parameters).

dpkt/ssl_ciphersuites.py
gendx: I will change it.

dpkt/ssl_ciphersuites.py
obormot: Makes sense, thanks. However what do you think about making the arguments optional (except "code")? I'm thinking of a use case where SSL decoder parses a packet and goes over a list of cipher suites. It may want to present them as a list of CipherSuite objects. All it has from the packet data is a list of "codes", and that's how it would instantiate these objects.

gendx: Exactly, the way to do it is with the BY_CODE table. Btw the current implementation in `ssl.py` does not extract the list of cipher suites for `TLSClientHello` packets (it just computes the number of proposed cipher suites), but I could improve it in another PR.

brifordwylie:
Since @obormot approved.. I'm going to 'hit the button' :)

PR #309, opened 2016-11-29
SCTP Parser Enhancement

wmnsk:
# SCTP Parser Enhancement

Implemented more precise decoding feature for SCTP. Please review :-)

### Problem

Currently sctp.py can parse only;

* SCTP Common Headers
  * Source Port
  * Destination Port
  * Verification Tag
  * Checksum
* SCTP Chunk Headers
  * Chunk Type
  * Chunk Flags
  * Chunk Length

The problem here is that it's hard to reach values in upper layers like diameter, as some more header fields exist in DATA type of SCTP chunk.  

### What's done

This enhancement includes the parser for **ALL fields** in each type defined in [RFC4960](https://tools.ietf.org/html/rfc4960).! Now we can get the values in upper layer simply by writing like;

```python
if isinstance(ip.data, dpkt.sctp.SCTP):
  sctp = ip.data
  for i in xrange(len(sctp.chunks))
    sctpchunk = sctp.chunks[i]
      if isinstance(sctpchunk, dpkt.sctp.ChunkData):
        diameter = sctpchunk.data  # <= GOT IT! Let's start decoding with diameter.py!
      else:
        pass
```

Other types are also supported. For example, we can count the number of **HEARTBEAT** and **HEARTBEAT_ACK**.

```python
if isinstance(ip.data, dpkt.sctp.SCTP):
  sctp = ip.data
  hb_count = 0
  hba_count = 0
  for i in xrange(len(sctp.chunks))
    sctpchunk = sctp.chunks[i]
      if isinstance(sctpchunk, dpkt.sctp.ChunkHeartbeat):
        hb_count += 1
        print 'HB Found. HB INFO: %s' % binascii.hexlify(sctpchunk.hb_info)
      elif isinstance(sctpchunk, dpkt.sctp.ChunkHeartbeat):
        hba_count += 1
        print 'HBAck Found. HB INFO: %s' % binascii.hexlify(sctpchunk.hb_info)
      else:
        pass
```

### Changes

Implemented the subclasses for each type of chunks, referring [RFC4960](https://tools.ietf.org/html/rfc4960).  
The fields to be decoded are describled in the docstring of each subclass. Here's the example for **DATA** type.

```
EXAMPLE: Chunk Type DATA from RFC.

       |     Source Port Number        |     Destination Port Number   |
       |                      Verification Tag                         |
       |                           Checksum                            |
       |   Type = 0    | Reserved|U|B|E|    Length                     |  # ^ fields supported currently
       |                              TSN                              |  # v Implemented by this commit
       |      Stream Identifier S      |   Stream Sequence Number n    |
       |                  Payload Protocol Identifier                  |
       \                                                               \
       /                 User Data (seq n of Stream S)                 /
       \                                                               \

```

```python
# Subclass for type DATA.

class ChunkData(Chunk):
    """SCTP DATA Chunk.
    This Class is to handle the 'DATA' type of SCTP Chunk.
    Supported fields are;
     - Chunk type
     - Chunk flags
     - Chunk length
     - Transmission sequence number
     - Stream identifier
     - Stream sequence number
     - Payload protocol identifier

    TODO: Longer class information....

    Attributes:
        __hdr__: Generic Chunk Header fields of SCTP.
        __hdr_spec__: Type-specific Chunk Header fields of SCTP.
        TODO.
    """
    __hdr_spec__ = (
        ('tsn', 'I', 0),
        ('stream_id', 'H', 0),
        ('stream_seq', 'H', 0),
        ('proto_id', 'I', 0)
    )
    __hdr__ = Chunk.__hdr__ + __hdr_spec__

    def unpack(self, buf):
        super(ChunkData, self).unpack(buf)
        self.data = self.data[:self.len - self.__hdr_len__]

```

These subclasses are inherited from 'Chunk' class which has currently been implemented.  
`__hdr_spec__` is the type-specific fields(chunk headers), and appended to `__hdr__`.  
If the type has no additional header, `__hdr__` is exactly same as the one in Chunk class.

### NOTICE

* The test methods are done in the same manner as before to keep it simple, but not testing all the subclasses newly implemented, as the sample SCTP byte arrays are not enough.
* Packet creation has not implemented fully in this commit, which is the same in current implementation.


wmnsk:
Found some errors thanks to the CI tools :-(
I'll commit and add comment here again after I correct them all.

wmnsk:
Created packets(manually wrote hex) and implemented test for ECNE, CWR, and ERROR. Coverage decrease is now only -0.002%, which I think not too much(considering the lines of code added).

**Could someone review?**

obormot:
@wmnsk I noticed some coverage decrease was from lack of unit test in crc32c.py, so I added one. You could cherry-pick commit 2cd44f8 into your pull request and see if it helps. If not - no big deal.

dpkt/sctp.py
```diff
@@ -25,29 +27,80 @@
...
+MISSING_MANDATORY_PARAMETER = 2
+STALE_COOKIE_ERROR = 3
+OUT_OF_RESOURCE = 4
+UNRESOLVABLE_ADDRESS = 5
+UNRECOGNIZED_CHUNK_TYPE = 6
+INVALID_MANDATORY_PARAMETER = 7
+UNRECOGNIZED_PARAMETER = 8
+NO_USER_DATA = 9
+COOKIE_RECEIVED_WHILE_SHUTTING_DOWN = 10
+RESTART_OF_AN_ASSOCIATION_WITH_NEW_ADDRESSES = 11
```
obormot: this can be made shorter

dpkt/sctp.py
```diff
@@ -4,9 +4,11 @@
...
 import crc32c
+import struct
+import binascii
 
 # Stream Control Transmission Protocol
-# http://tools.ietf.org/html/rfc2960
+# http://tools.ietf.org/html/rfc4960
 
 # Chunk Types
 DATA = 0
```
obormot: I'd add a common prefix to these constants, e.g.
```
CHUNK_TYPE_DATA = 0
CHUNK_TYPE_INIT = 1
...
```

and same for the error codes below
```
CHUNK_ERROR_INVALID_STREAM_IDENTIFIER = 1
...
```

dpkt/sctp.py
```diff
@@ -25,29 +27,80 @@
...
+        dpkt.Packet.unpack(self, buf)
+        return struct.unpack('B', self.data[0:1])[0]
+
     def unpack(self, buf):
+        self.ctype = self.get_chunktype(buf)
         dpkt.Packet.unpack(self, buf)
         l = []
         while self.data:
-            chunk = Chunk(self.data)
+            if self.ctype == DATA:
```
obormot: a map would be more elegant here, e.g.
```
CHUNK_TYPES_MAP = {
    CHUNK_TYPE_DATA: ChunkData,
    CHUNK_TYPE_INIT: ChunkInit,
   ...
}
```

then you can simply do 
```
chunk = CHUNK_TYPES_MAP.get(self.ctype, Chunk)(self.data)
```

dpkt/sctp.py
```diff
@@ -66,6 +119,18 @@ def __str__(self):
...
+    This Class is used as the SuperClass of all types of chunk and
+    as the Class of some chunks whose type is unknown or not implemented.
+    This only handles the three common part of Generic Chunk Headers;
+     - Chunk type
+     - Chunk flags
+     - Chunk length
+
+    Attributes:
+        __hdr__: Generic Chunk Header fields common in all types.
+        TODO.
```
obormot: (here and other classes below) you already described the attributes, so the TODO can go away

dpkt/sctp.py
```diff
@@ -77,28 +142,597 @@ def unpack(self, buf):
...
+    """
+    __hdr__ = Chunk.__hdr__
+
+    def unpack(self, buf):
+        super(ChunkShutdownComplete, self).unpack(buf)
+        self.data = self.data[:self.len - self.__hdr_len__]
+
+
+test_bytearr = {
```
obormot: if these are byte arrays it's a good idea to use the ```b'...'``` notation (also good for py3 compatibility)

dpkt/sctp.py
```diff
@@ -77,28 +142,597 @@ def unpack(self, buf):
...
+            assert (sctpchunk.num_outstream == 10)
+            assert (sctpchunk.num_instream == 10)
+            assert (sctpchunk.init_tsn == 3461729935)
+            assert (sctpchunk.cookie_param == 7)
+            assert (sctpchunk.param_len == 432)
+            assert (binascii.hexlify(sctpchunk.state_cookie) == 'bb7b0f9c0000fa00000a0000bb7b0f9c000000000fe4000001000000040000000101010a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000ce55ce8f0000fa00000a0000ce55ce8f0f1c000001000000040000000101011400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004f39d8584f3ac2b8')
```
obormot: these lines are way too long. you can break them up as follows:
```
assert (binascii.hexlify(sctpchunk.state_cookie) == ( 
    'bb7b0f9c0000fa00000a0000bb7b0f9c000000000fe400000100000'  # each line <= 100 chars
    '0040000000101010a00000000000000000000000000000000000000'
    ...
)

**Comment by wmnsk** [comment]
@obormot 
Thank you! Cherry-picked your crc32c update and followed your great insights.
Regarding the other changes made in the latest commit, I'll add comment for each part shortly.

File: dpkt/sctp.py [hunk 7]
```diff
@@ -5,6 +5,7 @@
 import dpkt
 import crc32c
 import struct
+import binascii
```
wmnsk: Removed because this is used only in testing part(importing in test method instead).

dpkt/sctp.py
```diff
@@ -421,82 +422,48 @@ class ChunkError(Chunk):
...
     __hdr__ = Chunk.__hdr__ + __hdr_spec__
 
-    @property
-    def error_spec(self, data, cause):
-        if cause == INVALID_STREAM_IDENTIFIER:
-            self.invalid_id = data[:self.len - self.__hdr_len__]
-            return 0
-        elif cause == MISSING_MANDATORY_PARAMETER:
+
+    def error_spec(self, data):
```
wmnsk: Removed unreasonable decorator(`@property`).

dpkt/sctp.py
```diff
@@ -167,6 +131,30 @@ class ChunkData(Chunk):
     )
     __hdr__ = Chunk.__hdr__ + __hdr_spec__
 
+    @property
```
wmnsk: Added more precise flags decoding in `INIT`.
This bitwise operations are inspired by the codes in [diameter.py](https://github.com/kbandla/dpkt/blob/master/dpkt/diameter.py#L42).

dpkt/sctp.py
```diff
@@ -188,20 +176,19 @@ class ChunkInit(Chunk):
...
         __hdr__: Generic Chunk Header fields of SCTP.
         __hdr_spec__: Type-specific Chunk Header fields of SCTP.
-        TODO.
     """
     __hdr_spec__ = (
         ('init_tag', 'I', 0),
         ('a_rwnd', 'I', 0),
-        ('num_outstream', 'H', 0),
-        ('num_instream', 'H', 0),
+        ('num_os', 'H', 0),
```
wmnsk: Modified the abbreviation following RFC.

dpkt/sctp.py
```diff
@@ -427,35 +407,35 @@ class ChunkError(Chunk):
 
 
     def error_spec(self, data):
-        if self.cause_code == INVALID_STREAM_IDENTIFIER:
-            self.invalid_id = struct.unpack('H', data[:self.len - self.__hdr_len__ -6])[0] >> 8
-        elif self.cause_code == MISSING_MANDATORY_PARAMETER:
+        if self.cause_code == CHUNK_ERR_INVALID_ID:
+            self.invalid_id = data[:self.len - self.__hdr_len__ -6]
```
wmnsk: Removed unpacking here. This error would be passed as binary like other errors.

dpkt/sctp.py
```diff
@@ -4,50 +4,69 @@
...
         ('sum', 'I', 0)
     )
 
+    def get_chunktype(self, buf):
+        dpkt.Packet.unpack(self, buf)
+        return struct.unpack('B', self.data[0:1])[0]
+
     def unpack(self, buf):
+        self.ctype = self.get_chunktype(buf)
         dpkt.Packet.unpack(self, buf)
```
obormot: calling dpkt.Packet.unpack(self, buf) twice here, first inside  get_chunktype(), then inside unpack(). Also not checking if data len is >= 1 before calling struct.unpack, potentially can cause an exception. I'd remove get_chunktype completely, and assign self.ctype inside unpack:
```
def unpack(self, buf):
    dpkt.Packet.unpack(self, buf)
    if self.data:
        self.ctype = struct.unpack('B', self.data[0])[0]
    l = []
    while self.data:
        ...
```

dpkt/sctp.py
```diff
@@ -77,28 +107,698 @@ def unpack(self, buf):
...
+
+    @property
+    def ending_flag(self):
+        return self.flags & 0x1
+
+    @ending_flag.setter
+    def ending_flag(self, e):
+        self.flags = (flags & ~0x1) | (e & 0x1)
+
+    def unpack(self, buf):
```
obormot: This method is duplicated as-is in many Chunk subclasses, is this really needed?  The parent class already has the same exact method, so maybe it'll all work automagically without duplicating the code?

wmnsk:
Thank you, @obormot! I revised the code following your review.
Regarding the flags, I had thought it's better to be able to access each bit directly by the name, but I deleted it this time as it was certainly too verbose.

obormot:
@wmnsk I see you addressed one of the comments, thanks for that! How about the 2nd comment, duplicate unpack() methods in almost every subclass of Chunk? Have you tried removing them in subclasses?

wmnsk:
@obormot
Oh sorry I misunderstood the point. I'll check my code again and get back to you. Thanks!

wmnsk:
@obormot 
Sure that the 'unpack' method should not be overridden, so I deleted them. In some of the subclasses, implemented the '@property' instead, to explicitly set the values in some optional fields to be parsed from 'self.data'.

dpkt/sctp.py
```diff
@@ -66,39 +82,684 @@ def __str__(self):
...
+            return self.data[:self.len - self.__hdr_len__]
+
+    @property
+    def _invalid_param(self):
+        # This should be None as no value would be inserted here.
+        if self.cause_code == CHUNK_ERR_INVALID_MAND_PARAM:
+            return self.data[:self.len - self.__hdr_len__]
+
+    @property
+    def _invalid_param(self):
```
obormot: lines 392 and 398 both define _invalid_param

dpkt/sctp.py
```diff
@@ -66,39 +82,684 @@ def __str__(self):
...
+        if self.cause_code == CHUNK_ERR_UNRECOG_PARAM:
+            return self.data[:self.len - self.__hdr_len__]
+
+    @property
+    def no_data_tsn(self):
+        if self.cause_code == CHUNK_ERR_NO_USER_DATA:
+            return self.data[:self.len - self.__hdr_len__]
+
+    @property
+    def _cookie_while_shut(self):
```
obormot: why properties prefixed with underscore?  properties are supposed to be public...

obormot:
@wmnsk There's nothing wrong with overriding the unpack() method when it provides customization of the subclass. But when it just copied as-is from the base class then it's redundant. I didn't mean to suggest removing all unpack methods, just to be clear!

wmnsk:
@obormot 
Thank you! I deleted underscored properties. These were ones whose values should be None because that kind of error has no fields over 'Cause Code' and 'Cause Length', but there's no reasonable reason to leave it explicitly.

wmnsk:
Sorry for pushing so often. I revised some mistake in the previous commit.

dpkt/sctp.py
kbandla: this should be `unrecog_param`

kbandla:
Awesome PR @wmnsk ! , and great comments from @obormot 
Packet format headers look good. The only thing was the `unrecog_param` property for `ChunkError`, but @obormot already pointed that out 👍 

Lastly, if you have the pcap for this, would you mind adding it to this issue? We might eventually have a test pcap corpus for the project. Thanks!

wmnsk:
Hi @kbandla,
Could you use pcaps available on [SampleCaptures page](https://wiki.wireshark.org/SampleCaptures#Stream_Control_Transmission_Protocol_.28SCTP.29) by wireshark?

For PCAP testing, I wrote a snippet on [Gist](https://gist.github.com/wmnsk/672d8384d34f4e5e0816d20029881aff). 

```shell-session
# dpktSCTP_testbypcap.py <PCAP>
[PACKET #1]
  === SCTP Common Header Info ===
  SCTP sport     :       32836
  SCTP dport     :       80
  SCTP vtag      :       0
  SCTP chksum    :       817557332

  [Chunk #0]
    === SCTP General Chunk Header Info ===
    Chunk type   :       1
    Chunk flags  :       0
    Chunk len    :       60

    === SCTP Type-Specific Chunk Info ===
    InitTag     :        1002019910
    A_RWND      :        106496
    OutstreamNo :        10
    InstreamNo  :        65535
    InitTSN     :        724401842

===========================================================

[PACKET #2]
  === SCTP Common Header Info ===
  SCTP sport     :       80
  SCTP dport     :       32836
  SCTP vtag      :       1002019910
  SCTP chksum    :       2746878182

  [Chunk #0]
    === SCTP General Chunk Header Info ===
    Chunk type   :       2
    Chunk flags  :       0
    Chunk len    :       232

    === SCTP Type-Specific Chunk Info ===
    InitTag     :        3530211813
    A_RWND      :        106496
    OutstreamNo :        10
    InstreamNo  :        10
    InitTSN     :        1677732374
    CookieParam :        7
    ParamLen    :        196
    StateCookie :        b3493015e1c27625f53ab8185555b8d...

===========================================================

[PACKET #3]
  === SCTP Common Header Info ===
  SCTP sport     :       32836
  SCTP dport     :       80
  SCTP vtag      :       3530211813
  SCTP chksum    :       1964287130

  [Chunk #0]
    === SCTP General Chunk Header Info ===
    Chunk type   :       10
    Chunk flags  :       0
    Chunk len    :       196

    === SCTP Type-Specific Chunk Info ===
    Cookie :     b3493015e1c27625f53ab8185555b8d...

===========================================================
```

Hope this helps :-)

PR #310, opened 2016-11-29
Added basic TLS extensions parsing in ssl.py

smutt:
This adds a function parse_extensions(buf) which can be called with a buffer starting with the TLS extensions 2 byte length TLV. So far I am only calling this function in TLSClientHello.unpack() but I think it is general enough to be called from any of the TLS packet parsing classes.

Hopefully I did all the github stuff properly. If not feel free to give me a kick :)


dpkt/ssl.py
```diff
@@ -250,6 +269,8 @@ def unpack(self, buf):
         self.num_compression_methods = parsed - 1
         self.compression_methods = map(ord, compression_methods)
         # extensions
+        self.extensions_length = struct.unpack('!H', self.data[pointer:pointer+2])[0]
```
obormot: why a separate field for the length, wouldn't ```len(self.extensions)``` give the same value?

dpkt/ssl.py
```diff
@@ -154,6 +154,25 @@ def parse_variable_array(buf, lenbytes):
...
 
 
+def parse_extensions(buf):
+    """
+    Parse TLS extensions in passed buf. Returns a dictionary of extensions with
+    ordinal extension type as key of dictionary and extension data as corresponding value.
+    Passed buf must start with the 2-byte extensions length TLV.
+    http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml
+    """
+    extensions_length = struct.unpack('!H', buf[:2])[0]
```
obormot: if buf is shorter than 2 this will throw an exception. same applies to line 169

dpkt/ssl.py
```diff
@@ -154,6 +154,25 @@ def parse_variable_array(buf, lenbytes):
     return data, size + lenbytes
 
 
+def parse_extensions(buf):
+    """
+    Parse TLS extensions in passed buf. Returns a dictionary of extensions with
```
obormot: what if the same extension type appears more than once? also, dict is not ordered. in other dpkt parsers fields like this are usually returned as list of tuples, such as [(ext1_type, ext1_value), (ext2_type, ext2_value), ...]

dpkt/ssl.py
```diff
@@ -154,6 +154,25 @@ def parse_variable_array(buf, lenbytes):
...
+    ordinal extension type as key of dictionary and extension data as corresponding value.
+    Passed buf must start with the 2-byte extensions length TLV.
+    http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml
+    """
+    extensions_length = struct.unpack('!H', buf[:2])[0]
+    extensions = {}
+
+    pointer = 2
+    while(pointer < extensions_length):
```
obormot: parenthesis ( ) are redundant

smutt:
On 11/29/16 22:49, Oscar wrote:
>
> *@obormot* requested changes on this pull request.
>
> ------------------------------------------------------------------------
>
> In dpkt/ssl.py 
> <https://github.com/kbandla/dpkt/pull/310#pullrequestreview-10687066>:
>
> > @@ -250,6 +269,8 @@ def unpack(self, buf):
>           self.num_compression_methods = parsed - 1
>           self.compression_methods = map(ord, compression_methods)
>           # extensions
> +        self.extensions_length = struct.unpack('!H', self.data[pointer:pointer+2])[0]
>
> why a separate field for the length, wouldn't |len(self.extensions)| 
> give the same value?
self.extensions_length is the length in bytes of all the extensions.
len(self.extensions) would return the number of items in the Python data 
dictionary. They're different.

We actually don't need self.extensions_length at all and can remove it. 
Would this resolve this issue?

> ------------------------------------------------------------------------
>
> In dpkt/ssl.py 
> <https://github.com/kbandla/dpkt/pull/310#pullrequestreview-10687066>:
>
> > @@ -154,6 +154,25 @@ def parse_variable_array(buf, lenbytes):
>       return data, size + lenbytes
>   
>   
> +def parse_extensions(buf):
> +    """
> +    Parse TLS extensions in passed buf. Returns a dictionary of extensions with
> +    ordinal extension type as key of dictionary and extension data as corresponding value.
> +    Passed buf must start with the 2-byte extensions length TLV.
> +    http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml
> +    """
> +    extensions_length = struct.unpack('!H', buf[:2])[0]
>
> if buf is shorter than 2 this will throw an exception. same applies to 
> line 169
>
I doubt you'll ever encounter a TLS Handshake packet in the wild that 
has no extensions. It is not permitted by protocol, and it would be 
impossible to establish a TLS connection without the use of extensions 
in the Handshake packet.

If you'd like I can add code to test that there are extensions before 
parsing them. If that would resolve this.

It is possible for an extension to have data of zero length. But 
parse_variable_array() should handle that.
> ------------------------------------------------------------------------
>
> In dpkt/ssl.py 
> <https://github.com/kbandla/dpkt/pull/310#pullrequestreview-10687066>:
>
> > @@ -154,6 +154,25 @@ def parse_variable_array(buf, lenbytes):
>       return data, size + lenbytes
>   
>   
> +def parse_extensions(buf):
> +    """
> +    Parse TLS extensions in passed buf. Returns a dictionary of extensions with
>
> what if the same extension type appears more than once? also, dict is 
> not ordered. in other dpkt parsers fields like this are usually 
> returned as list of tuples, such as [(ext1_type, ext1_value), 
> (ext2_type, ext2_value), ...]
>
Section 7.4.1.4 of RFC 5246 says:
When multiple extensions of different types are present in the
ClientHello or ServerHello messages, the extensions MAY appear in any
order.  There MUST NOT be more than one extension of the same type.

I checked the RFCs that update RFC 5246 and this text is still normative.

Given the above I think using a Python dict makes sense.

smutt:
On 11/29/16 23:00, Oscar wrote:
>
> *@obormot* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In dpkt/ssl.py 
> <https://github.com/kbandla/dpkt/pull/310#pullrequestreview-10688007>:
>
> > @@ -154,6 +154,25 @@ def parse_variable_array(buf, lenbytes):
>       return data, size + lenbytes
>   
>   
> +def parse_extensions(buf):
> +    """
> +    Parse TLS extensions in passed buf. Returns a dictionary of extensions with
> +    ordinal extension type as key of dictionary and extension data as corresponding value.
> +    Passed buf must start with the 2-byte extensions length TLV.
> +    http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml
> +    """
> +    extensions_length = struct.unpack('!H', buf[:2])[0]
> +    extensions = {}
> +
> +    pointer = 2
> +    while(pointer < extensions_length):
>
> parenthesis ( ) are redundant
Good catch! I've updated my code. I'll wait for the other issues you 
raised to be resolved before doing any git pushing/pulling.

obormot:
> I doubt you'll ever encounter a TLS Handshake packet in the wild that
has no extensions. It is not permitted by protocol, and it would be
impossible to establish a TLS connection without the use of extensions
in the Handshake packet.

The idea is that dpkt must be resilient to malformed packets. 

> Section 7.4.1.4 of RFC 5246 says:
When multiple extensions of different types are present in the
ClientHello or ServerHello messages, the extensions MAY appear in any
order.  There MUST NOT be more than one extension of the same type.

Again, we can't just blindly trust RFC when parsing packets. They can be malformed. Pcap files can be cut in the middle etc. 

> Given the above I think using a Python dict makes sense.

Just trying to keep it consistent with other parsers (I think DHCP options is an example). It's trivial to convert an ordered list of tuples to a dict, just be calling dict() on it, if someone needs it in their app implementation. However it's impossible to convert it in reverse while maintaining order and keeping duplicates intact.

smutt:
I have committed some changes that should address all comments received on this pull-request. Please let me know if this resolves the issues.

dpkt/ssl.py
```diff
@@ -154,6 +154,26 @@ def parse_variable_array(buf, lenbytes):
...
+    """
+    extensions_length = struct.unpack('!H', buf[:2])[0]
+    extensions = []
+
+    pointer = 2
+    while pointer < extensions_length:
+        ext_type = struct.unpack('!H', buf[pointer:pointer+2])[0]
+        pointer += 2
+        ext_data, parsed = parse_variable_array(buf[pointer:], 2)
+        extensions.append([ext_type, ext_data])
```
obormot: this appends lists, not tuples

dpkt/ssl.py
```diff
@@ -249,7 +269,10 @@ def unpack(self, buf):
         pointer += parsed
         self.num_compression_methods = parsed - 1
         self.compression_methods = map(ord, compression_methods)
-        # extensions
+        # There should always be atleast 1 extension
+        if struct.unpack('!H', self.data[pointer:pointer+2])[0] < 6:
```
obormot: I don't think extensions are required (in rfc). If there's data following the compression methods then it must be extensions. So I recommend something like this instead:
```
if self.data[pointer:]: 
    self.extensions = parse_extensions(self.data[pointer:])
```

or (better) check for a minimal length of at least 1 extension:
```
if len(self.data[pointer:]) >= ... :
    self.extensions = parse_extensions(self.data[pointer:])
```

smutt:
Made changes to resolve issues. Again thanks for taking time to review these.

kbandla:
lgtm

smutt:
Can you merge this soon please? I'm hoping to do some more work on ssl.py soon and it would be much easier if this PR could be merged first before I create another. Thanks!

PR #311, opened 2016-12-04
Fix loop condition in tls_multi_factory.

gendx:
I fixed a bug in the loop condition of `tls_multi_factory`, that caused an exception for too small inputs (even if the input is a prefix of a well-formed TLS record).


gendx:
I made this compatible with the new Python 3 upgrade. Could anyone else review ?

gendx:
PING @brifordwylie @kbandla 
I didn't open an issue for this, but this is actually a bug that breaks my code in practice. The fix is only one line and I added the tests. Could you please merge this ?

PR #312, opened 2016-12-20
Added TLS and TLSCertificate classes

smutt:
The addition of these two classes is backwards compatible with the previous ssl.py. I have tested this a bunch with an application I am developing. Let me know if there is anything you would like to change.

Thanks,
Andrew


obormot:
@smutt please add some unit tests for the new classes

smutt:
Test cases added.

dpkt/ssl.py
```diff
@@ -178,6 +178,24 @@ class SSL3Exception(Exception):
...
+    def __init__(self, *args, **kwargs):
+        self.records = []
+        dpkt.Packet.__init__(self, *args, **kwargs)
+        self.length = len(self.data)
+
+
+    def unpack(self, buf):
+        dpkt.Packet.unpack(self, buf)
+        pointer = 0
+        while len(self.data[pointer:]) > 0:
```
obormot: A couple of issues here:

1. You're moving the pointer and parsing self.data chunks into self.records. However self.data remains the same (i.e. you're missing something like ```self.data = self.data[pointer:]``` in each iteration). 
2. Missing the packing operation, and a unit test for it (e.g. ```assert str(TLS('foo')) == 'foo'```). You'd need to address the first issue for the packing to work correctly.

smutt: Thanks for the fast response. I'm going to hold on this PR until I write the code in my application that actually uses the certificates. That should highlight any problems with this code more easily. I'll get back to this PR in a week or two.

smutt: Hi Obormot,

I'm revisiting this after mostly finishing the application that actually uses this code. Thanks for being patient.

I'm confused about your request to mutate self.data. Is there a requirement to reduce self.data that's indepedent of this class? I ask because this code works, but I understand there may be a standard requirement in dpkt that self.data be reduced as it is being consumed.

I'm also confused about the need to create a pack() function as I don't see any other pack() functions in this file. I have created a test case for classes TLS and TLSCertificate.

Thanks again for being patient.

obormot: @smutt 
> there may be a standard requirement in dpkt that self.data be reduced as it is being consumed
Yes, exactly correct. self.data is whatever bytes remain after parsing a protocol layer. 

> confused about the need to create a pack() function as I don't see any other pack() functions in this file
Custom pack functions are only needed when standard packing does not work (i.e. the unit test I've given as an example fails). I was simply speculating that if your code does not consume data as it parses it, then the data may appear twice in the packed buffer, which would obviously be incorrect

smutt: Please review my latest commit and advise.

smutt:
Unless there is something else that I need to add for this to be merged please merge it.

brifordwylie:
@smutt Since @obormot did the review he'll have to Approve before we merge. Thanks for being patient, we're all supporting this project 'around the edges' :)

obormot:
I will look into this tonight @smutt @brifordwylie

dpkt/ssl.py
obormot: Hi @smutt 
It feels strange to see the TLS class having a different ```__hdr__``` compared to SSL2, essentially packet structure is the same, so why are the fields different?  Secondly, you're creating the ```length``` field as ```len(self.data)```, then you're reducing the size of data but length stays the same. While this approach may work for your use case, it should be following the dpkt style more closely to become part of the library, IMO.  In general looking at ssl.py though, it is a bit of a mess and could use some refactoring.

smutt:
Thanks for looking at this Oscar. Let me spend some time refactoring this and come up with a better, more general, solution. I really appreciate your time in reviewing this.

smutt:
Let's try this again.
This version should be a bit more dpkt'ish. The SSL/TLS protocol has changed significantly since SSL 2, so I left the SSL2 class alone. Instead the TLS class can be used for anything SSL 3.0 upto TLS 1.2. I also actually implemented the header and removed the creation of length in __init__().

I have tested this with SSL 3.0, TLS 1.0 and TLS 1.2. I couldn't find any TLS 1.1 packets to test with.

smutt:
Could someone please take a look at this and hopefully merge it? I believe I've incorporated all suggestions.

I've built a LEDE/OpenWRT package for python-dpkt that I plan to push to their packages repository soon, but currently I'm using my fork for that. I can't start using this main fork until I get this functionality merged.

Again, many thanks to everyone who helped me with this. I understand we all have real jobs, too.

brifordwylie:
@smutt Thanks for being so patient, the code looks good to me but I'm going to let @obormot check off on the details. I know he's even more busy lately than usually but perhaps if we call him 3 times he'll magically appear and approve @obormot @obormot @obormot  :)

obormot:
@smutt @brifordwylie I'm terribly sorry it has been sitting in my queue for a week. The change looks good to me, the only thing I noticed is that there's a field named "length" in the new TLS class, while the same field was named "len" in existing SSL classes. These should be made consistent, but otherwise everything looks good.

smutt:
Thanks Oscar. I changed attribute TLS.length to TLS.len so it is consistent with the SSL class.

brifordwylie:
@smutt do you need a new pypi release after this merge?

smutt:
The LEDE/OpenWRT package Makefiles take the code directly from github. They identify releases by git commit hashes, so no PyPi release needed.

PR #313, opened 2016-12-26
Diameter Enhancement: support padding in AVPs

wmnsk:
# Diameter enhancement: Support for padding in AVPs

Hi,  
I implemented the padding support in diameter AVPs. Please review :-)

### Problem in current implementation

diameter.py fails parsing multiple AVPs due to the lack of consideration for padding.  
When an AVP has its data with padding, `AVP` class can only parse actual data part referring the AVP Length field, and it results in the failure of parsing the next AVP(AVP Code is parsed as zero...).

```
      |                           AVP Code                            |
      |V M P r r r r r|                  AVP Length                   |
      |                        Vendor-ID (opt)                        |
      |    Data ...       <= This sometimes has padding, which is not counted in Length.
      +-+-+-+-+-+-+-+-+
```

### What's done

Added the `padlen` instance variable and let the `data` include padding with the following steps.

1. Checks the actual AVP Length(self.len) if it is a multiple of four.
    * If so, `padlen` will be 0
    * If not, `padlen` is set to 1-3, depending on how many length is needed to fill.
2. Adds the `padlen` to the length of `data`. 

### Other modifications

These don't affect the parsing logic, but I added some lines of code for enhancement.

* Wrote Class information docstrings.
* Let the command codes inside the `Diameter` class, as they won't be referred from other classes.
* Added `avp_codes` in `AVP` class(Only supports the ones defined in rfc6733)
* Added test bytearrays to test all patterns of padding length.

### Further enhancements

This is not included in this commit but I'm planning to implement some application-specific AVPs like Gx/Gy in other class(or even other file maybe). Let me create a new issue to discuss how we should implement it.

### DIFF
--- a/dpkt/ssl.py
+++ b/dpkt/ssl.py
@@ -154,6 +154,26 @@ def parse_variable_array(buf, lenbytes):
     return data, size + lenbytes
 
 
+def parse_extensions(buf):
+    """
+    Parse TLS extensions in passed buf. Returns an ordered list of extension tuples with
+    ordinal extension type as first value and extension data as second value.
+    Passed buf must start with the 2-byte extensions length TLV.
+    http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml
+    """
+    extensions_length = struct.unpack('!H', buf[:2])[0]
+    extensions = []
+
+    pointer = 2
+    while pointer < extensions_length:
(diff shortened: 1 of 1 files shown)


wmnsk:
Closed this PR as I found `pack_hdr` also needs padding consideration...
I've created **[new PR](https://github.com/kbandla/dpkt/pull/314)**. Please take a look at it instead. Thanks.

PR #314, opened 2016-12-27
Diameter Enhancement: support padding in AVP

wmnsk:
# Diameter enhancement: Support padding in AVPs

Hi,  
I implemented the padding support in diameter AVPs. Please review :-)

### Problem in current implementation

diameter.py fails parsing multiple AVPs due to the lack of consideration for padding.  
When an AVP has its data with padding, `AVP` class can only parse actual data part referring the AVP Length field, and it results in the failure of parsing the next AVP(AVP Code is parsed as zero...).

```
      |                           AVP Code                            |
      |V M P r r r r r|                  AVP Length                   |
      |                        Vendor-ID (opt)                        |
      |    Data ...       <= This sometimes has padding, which is not counted in Length.
      +-+-+-+-+-+-+-+-+
```

### What's done
Added padding feature when **packing Diameter header** and **unpacking Diameter AVP**.

#### Changes in Pack Diameter Header
1. Checks the each AVP(retrieved from `self.data`) Length if it is a multiple of four.
    * If so, `padlen` will be 0
    * If not, `padlen` is set to 1-3, depending on how much length is needed to fill.
2. Depending on `padlen`, creates null bytearray(varies from `b''` to `b'\x00\x00\x00'`) and append it to AVP.
3. Joins each AVPs as `self.data`.

**Example:**

Suppose Diameter class got AVP1 and AVP2, whose actual length is 9 and 18.

```
+-+-+-+-+-+-+-+-+-+
|  AVP1  (len=9)  |
+-+-+-+-+-+-+-+-+-+
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|           AVP2 (len=18)           |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

To adjust their length to multiple of four, add padding for each AVP.  

```
+-+-+-+-+-+-+-+-+-+-+-+-+
|  AVP1  (len=9)  | p1  |
+-+-+-+-+-+-+-+-+-+-+-+-+
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|           AVP2 (len=18)           |p2 |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

Now their length is 12 and 20. So let's merge them as a set of AVPs(=data of diameter header).

```
|  AVP1  (len=9)  | p1  |           AVP2 (len=18)           |p2 |
```

#### Changes in Unpack AVP
1. Checks the actual AVP Length(self.len) if it is a multiple of four.
    * If so, `padlen` will be 0
    * If not, `padlen` is set to 1-3, depending on how much length is needed to fill.
2. Adds the `padlen` to the length of `data`. 

**Example:**

An AVP has its length in Length field. If that length is not multiple of four, it means it hes the padding to fill.  
When comparing the length value the actual length. The gap between them is the length of padding.

```
/ value in length \ 
/actual length of bytes \
|  AVP1  (len=9)  |     |           AVP2 (len=18)           |p2 |
```

So let's unpack with actual length by adding the length of padding to the target range of unpacking, not to parse padding bytes as next AVP.

```
/range to unpack as AVP1\
|  AVP1  (len=9)  |     |           AVP2 (len=18)           |p2 |
```

If not doing this, AVP2 cannot be parsed correctly.

```
                  /      wrong range to unpack as AVP2         \
|  AVP1  (len=9)  |     |           AVP2 (len=18)           |p2 |
                   ^^^^^
                   AVP Code would be zero, parsing results in failure.
```

### Other modifications

These don't affect the creating/parsing logic, but I added some lines of code for enhancement.

* Wrote Class information docstrings.
* Let the command codes inside the `Diameter` class, as they won't be referred from other classes.
* Added `avp_codes` in `AVP` class(Only supports the ones defined in rfc6733)
* Modified test method to make it simple. Bytearray for testing is no longer needed.

### Further enhancements

This is not included in this commit but I'm planning to enhance this further. Let me create a new issue in the future to discuss how we should implement them if needed.

1. Implement child classes for basic AVPs(where app_id in header is 0) to create/parse it in more easy way.
2. Implement some application-specific AVPs like Gx/Gy in other class(or even other file maybe).

### DIFF
--- a/dpkt/diameter.py
+++ b/dpkt/diameter.py
@@ -6,29 +6,22 @@ import struct
 import dpkt
 from decorators import deprecated
 
-# Diameter Base Protocol - RFC 3588
-# http://tools.ietf.org/html/rfc3588
-
-# Request/Answer Command Codes
-ABORT_SESSION = 274
-ACCOUTING = 271
-CAPABILITIES_EXCHANGE = 257
-DEVICE_WATCHDOG = 280
-DISCONNECT_PEER = 282
-RE_AUTH = 258
-SESSION_TERMINATION = 275
+# Diameter Base Protocol - RFC 6733
+# http://tools.ietf.org/html/rfc6733
 
 
 class Diameter(dpkt.Packet):
-    """Diameter.
-
-    TODO: Longer class information....
+    """Diameter Base Protocol Header.
+    This class is to parse diameter header and split AVPs above.
 
     Attributes:
         __hdr__: Header fields of Diameter.
-        TODO.
+        cmd_codes: Basic diameter command codes defined in rcf6733.
+                   Other commands should be implemented by overridding
+                   this attribute from another class specific to
+                   each application.
     """
-    
+
     __hdr__ = (
         ('v', 'B', 1),
         ('len', '3s', 0),
@@ -38,6 +31,16 @@ class Diameter(dpkt.Packet):
         ('hop_id', 'I', 0),
         ('end_id', 'I', 0)
     )
+
+    cmd_codes = {
+        'ABORT_SESSION': 274,
+        'ACCOUTING': 271,
+        'CAPABILITIES_EXCHANGE': 257,
+        'DEVICE_WATCHDOG': 280,
+        'DISCONNECT_PEER': 282,
+        'RE_AUTH': 258,
+        'SESSION_TERMINATION': 275
+    }
 
     @property
     def request_flag(self):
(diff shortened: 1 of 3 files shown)


wmnsk:
Sorry for updating too frequently... I mis-committed changes.
I'm closing this PR and open another to avoid annoying git operations and keep the commit log clean.

Issue #324, opened 2017-01-29
ICMP payload inside IP packets isn't parsed under Python2


Tojaj:
Current code in dpkt git (1.9.0) cannot parse ICMP packets (and some other protocols) when run under Python2, but works fine under Python3.

My git HEAD is 65b3d9716b80d17e5bffb66f57230ddeee384f75

## Reproducer

    import dpkt
    with open("ip_frag_source.pcap", 'rb') as f:
        pcap = dpkt.pcap.Reader(f)
        for ts, buf in pcap:
            eth = dpkt.ethernet.Ethernet(buf)
            ip = eth.data
            print(type(ip.data))

Note: The PCAP file ``ip_frag_source.pcap`` can be downloaded from http://chrissanders.org/packet-captures/

### Python 2

    $ python2.7 reproducer.py
    <type 'str'>
    ....

### Python 3

    $ python3.5 reproducer.py
    <class 'dpkt.icmp.ICMP'>
    ....

## Debugging

I added raise statement into ``__load_protos`` function in ``ip.py`` to see the exception:

    try:
        mod = __import__(name, g, level=1)
        IP.set_proto(v, getattr(mod, name.upper()))
    except (ImportError, AttributeError) as err:
        if name == "icmp":  # XXX Debug
            raise           # XXX Debug
        continue

and result is:

    $ python2.7 ./reproducer.py 
    Traceback (most recent call last):
      File "./reproducer.py", line 3, in <module>
        import dpkt
      File "/home/tojaj/git/dpkt/dpkt/__init__.py", line 33, in <module>
        from . import icmp
      File "/home/tojaj/git/dpkt/dpkt/icmp.py", line 8, in <module>
        from . import ip
      File "/home/tojaj/git/dpkt/dpkt/ip.py", line 362, in <module>
        __load_protos()
      File "/home/tojaj/git/dpkt/dpkt/ip.py", line 354, in __load_protos
        IP.set_proto(v, getattr(mod, name.upper()))
    AttributeError: 'module' object has no attribute 'ICMP'

### Code

    import dpkt
    from pprint import pprint
    pprint(dpkt.ip.IP._protosw)

### Python2

    $ python2.7 ./reproducer.py 
    {0: <class 'dpkt.ip.IP'>,
     2: <class 'dpkt.igmp.IGMP'>,
     4: <class 'dpkt.ip.IP'>,
     6: <class 'dpkt.tcp.TCP'>,
[...]

PR #325, opened 2017-01-29
Do circular imports in icmp and icmp6 modules correctly (Issue #324)

Tojaj:
There are circular deps between ip.py and icmp.py/icmp6.py modules.
Because of alphabetical imports in __init__.py, the icmp modules
are imported before ip.py, but they both need ip.py and ip.py during
initialization tries to import these icmp.

Because of these circular imports, icmp and icmp6 data in IP packets
weren't parsed correctly (#324). This patch does the imports in
slightly better fashion, so parsing of ICMP packets in IP works
again.

### DIFF
--- a/dpkt/icmp.py
+++ b/dpkt/icmp.py
@@ -5,7 +5,6 @@ from __future__ import print_function
 from __future__ import absolute_import
 
 from . import dpkt
-from . import ip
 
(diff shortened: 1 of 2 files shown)


obormot:
Hi @Tojaj this makes sense and the fix looks good to me. While at it could you also look at fixing GRE?

PR #329, opened 2017-02-03
Do not parse L4 headers for fragmented IP packets with fragment offset > 0

### LINKED ISSUE #102: Don't try to decode ip packets with offset eg. fragments
_From [noro...@gmail.com](https://code.google.com/u/100262187313124309206/) on September 06, 2011 04:52:48_

What steps will reproduce the problem? 1. Dump fragmented udp traffic and it will some time decode the fragments with offset > 0 as UDP headers. What is the expected output? What do you see instead? Expected: Just IP pkt with data payload.
What we get: UDP headers when data is decodable. What version of the product are you using? On what operating system? SVN rev 82, Linux

**Attachment:** [dont-unpack-ip-fragments.patch](http://code.google.com/p/dpkt/issues/detail?id=75)

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=75_

KnifeMaster007:
Fixed #102 issue

### DIFF
--- a/dpkt/ip.py
+++ b/dpkt/ip.py
@@ -140,7 +140,7 @@ class IP(dpkt.Packet):
         else:  # very likely due to TCP segmentation offload
             buf = buf[self.__hdr_len__ + ol:]
         try:
-            self.data = self._protosw[self.p](buf)
+            self.data = self._protosw[self.p](buf) if self.offset == 0 else buf
             setattr(self, self.data.__class__.__name__.lower(), self.data)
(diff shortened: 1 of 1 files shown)


KnifeMaster007:
@obormot Thanks! But I've no idea, why Travis check fails on test_format_request in http.py.

obormot:
@KnifeMaster007 I restarted the build and it worked fine

KnifeMaster007:
@obormot Thanks again)

PR #335, opened 2017-03-09
Forces HTTP handling to use bytes to improve Python 3 compatibility

scott-hand:
Right now, http.py uses cStringIO to parse data. This cannot handle any Unicode data that can't be parsed as ASCII ([source](https://docs.python.org/2/library/stringio.html#cStringIO.StringIO)). Unfortunately, this creates a situation in which the encoding applied on the string passed into the dpkt classes differs between Python 2 and Python 3, making it harder to write flexible code for either.

### DIFF
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -373,6 +373,30 @@ def test_invalid_header():
         assert False
 
 
+def test_gzip_response():
+    import zlib
+    # valid response, compressed using gzip
+    s = b'HTTP/1.0 200 OK\r\n' \
+        b'Server: SimpleHTTP/0.6 Python/2.7.12\r\n' \
+        b'Date: Fri, 10 Mar 2017 20:43:08 GMT\r\n' \
+        b'Content-type: text/plain\r\n' \
+        b'Content-Encoding: gzip\r\n' \
+        b'Content-Length: 68\r\n' \
+        b'Last-Modified: Fri, 10 Mar 2017 20:40:43 GMT\r\n\r\n' \
+        b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x02\x03\x0b\xc9\xc8,V\x00\xa2D' \
+        b'\x85\xb2\xd4\xa2J\x85\xe2\xdc\xc4\x9c\x1c\x85\xb4\xcc\x9cT\x85\x92' \
+        b'|\x85\x92\xd4\xe2\x12\x85\xf4\xaa\xcc\x02\x85\xa2\xd4\xe2\x82\xfc' \
+        b'\xbc\xe2\xd4b=.\x00\x01(m\xad2\x00\x00\x00'
+    r = Response(s)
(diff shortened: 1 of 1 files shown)


scott-hand:
Added small tweak to address the fact that Python 3.4 has no byte interpolation. TIL.

scott-hand:
HTTP (at least in my opinion) needs to be parsed as bytes from a stream of binary data (BytesIO) and then, once processed, decoded. I would say that it absolutely needs to be processed as bytes because of cases like gzipped response data, non-HTML responses like images, etc. For many pages, the charset parameter in the Content-Type header will tell you to decode it to Unicode, and for others it's in <meta> tags.

So the tricky part is that you want to parse, for example, UTF8 pages and get a Unicode result from dpkt. However, you might want to parse a .jpg response and get bytes back. So IMO there are two solutions to this:

1. Try to figure out the encoding and decode it yourself.
    * Pros
        * It Just Works for most users
    * Cons
        * More code, somewhat at odds with dpkt's lightweight ethos (or at least my perception of it)
        * Have to potentially parse HTML to fine those <meta> tag with charset, or at least grep for it
        * Potentially undefined and surprising behavior when a Unicode page doesn't provide a charset
2. Return bytes and let user code figure it out
    * Pros
        * Less work for dpkt
        * Avoids wasting cycles for cases in which people don't care, can work with bytes
    * Cons
        * Might be surprising if people expected a str out of the body field

In my opinion, dpkt's goal is to be a *protocol parser* and while HTTP fields can be used to specify the charset encoding, I don't think that performing Unicode decoding should be coupled with HTTP parsing, so I favor the latter option. A compromise might be a feature down the road in which users can set a `decode=True` to ask dpkt to search for a charset and decode if one is found.

Regarding the header values and HTTP fields ("POST", "200", etc.), you're right that it "feel wrong" and it certainly brings out the ugliness of the Python b"..." bytes literal. HTTP have this to say about headers:

>In practice, most HTTP header field values use only a subset of the US-ASCII charset [USASCII]. Newly defined header fields SHOULD limit their field values to US-ASCII octets.  A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

([Source](https://tools.ietf.org/html/rfc7230#section-3.2.4))

So IMO, the thing to do here is to decode the HTTP fields and the headers as Unicode ASCII encoding with the `errors="ignore"` argument. I will do that and submit a new commit to this PR.

The data that causes issues is gzipped data, which isn't covered in the unit tests right now. I will generate some that causes the issue I'm describing and commit it as well.

scott-hand:
As the next commit would be completely reworking the original one, I'm gonna close this PR and rework the solution into something that matches more expected behavior (HTTP attributes as ASCII Unicode per the RFC and bytes returned in body)

brifordwylie:
@scott-hand your proposed solution sounds good to me. Please open a new PR when you're done and thanks again for all the effort on this.

PR #336, opened 2017-03-10
Python 3 support for bytes in and out of http module classes

scott-hand:
Creating a new PR to overhaul my previous one with a more structured approach to more fully supporting Python 3 in dpkt.http.

The problem currently is that, while many dpkt modules take bytes and return bytes in their class `body` properties, the http module currently expects a Unicode string due to its use of StringIO. Python 2 uses cStringIO if possible, which cannot handle Unicode data with non-ASCII characters. This results in a conflict in which Python 3 code must decode bytes to Unicode data (in "latin-1") for StringIO to work correctly, while Python 2 will encounter problems when attempting to run that same code. This is particularly apparent in gzip compressed responses and other byte streams, as is demonstrated in the `test_gzip_response()` test added to document this issue.

My approach is to do the following:

* Switch to using BytesIO. This requires that data be provided as bytes rather than str like it was previously. This is intended and desired behavior, as, like discussed previously, many HTTP requests or responses must handle binary data.
* Parse the headers and HTTP protocol fields as ASCII decoded strings from the bytes with errors in decoding ignored. This is done per RFC 7230 advice: "In practice, most HTTP header field values use only a subset of the US-ASCII charset [USASCII]. Newly defined header fields SHOULD limit their field values to US-ASCII octets.  A recipient SHOULD treat other octets in field content (obs-text) as opaque data." ([Source](https://tools.ietf.org/html/rfc7230#section-3.2.4))
* Store the `data` and `body` properties of the classes as bytes.
* Add a `__bytes__` method because, since classes take bytes now, code like `Request(str(r))` must become `Request(bytes(r))`. The `__str__` method still remains and attempts to decode bytes as utf8.

In my opinion, dpkt's goal is to be a protocol parser and while HTTP fields can be used to specify the charset encoding, I don't think that performing Unicode decoding is or should be coupled with HTTP parsing, so I chose for it to provide the output as bytes. Finding the correct charset involves both looking in headers as well as parsing HTML, so I would prefer to leave that as a high level operation done by users, but a compromise might be to add a `decode=True` option or something similar at some point in the future.

### DIFF
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -373,6 +373,30 @@ def test_invalid_header():
         assert False
 
 
+def test_gzip_response():
+    import zlib
+    # valid response, compressed using gzip
+    s = b'HTTP/1.0 200 OK\r\n' \
+        b'Server: SimpleHTTP/0.6 Python/2.7.12\r\n' \
+        b'Date: Fri, 10 Mar 2017 20:43:08 GMT\r\n' \
+        b'Content-type: text/plain\r\n' \
+        b'Content-Encoding: gzip\r\n' \
+        b'Content-Length: 68\r\n' \
+        b'Last-Modified: Fri, 10 Mar 2017 20:40:43 GMT\r\n\r\n' \
+        b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x02\x03\x0b\xc9\xc8,V\x00\xa2D' \
+        b'\x85\xb2\xd4\xa2J\x85\xe2\xdc\xc4\x9c\x1c\x85\xb4\xcc\x9cT\x85\x92' \
+        b'|\x85\x92\xd4\xe2\x12\x85\xf4\xaa\xcc\x02\x85\xa2\xd4\xe2\x82\xfc' \
+        b'\xbc\xe2\xd4b=.\x00\x01(m\xad2\x00\x00\x00'
+    r = Response(s)
+    assert r.version == b'1.0'
+    assert r.status == b'200'
+    assert r.reason == b'OK'
+    # Make a zlib compressor with the appropriate gzip options
+    decompressor = zlib.decompressobj(16 + zlib.MAX_WBITS)
+    body = decompressor.decompress(r.body)
+    assert body.startswith(b'This is a very small file')
+
+
 if __name__ == '__main__':
     # Runs all the test associated with this class/file
     test_parse_request()
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -3,6 +3,11 @@
 """Hypertext Transfer Protocol."""
 from __future__ import print_function
 from __future__ import absolute_import
+try:
+    from collections import OrderedDict
+except ImportError:
+    # Python 2.6
+    OrderedDict = dict
 
 from . import dpkt
 from .compat import BytesIO, iteritems
@@ -10,7 +15,7 @@ from .compat import BytesIO, iteritems
 
 def parse_headers(f):
     """Return dict of HTTP headers parsed from a file object."""
-    d = {}
+    d = OrderedDict()
     while 1:
         # The following logic covers two kinds of loop exit criteria.
         # 1) If the header is valid, when we reached the end of the header,
@@ -91,7 +96,7 @@ class Message(dpkt.Packet):
         if args:
             self.unpack(args[0])
         else:
-            self.headers = {}
+            self.headers = OrderedDict()
             self.body = b''
             self.data = b''
             # NOTE: changing this to iteritems breaks py3 compatibility
--- a/dpkt/http.py
+++ b/dpkt/http.py
@@ -5,7 +5,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 
 from . import dpkt
-from .compat import StringIO, iteritems
+from .compat import BytesIO, iteritems
 
 
 def parse_headers(f):
@@ -19,7 +19,7 @@ def parse_headers(f):
         # 2) If this is a weird header, which do not ends with '\r\n',
         #    f.readline() would return with '', then after strip(),
         #    we still get an empty string, also break the loop.
-        line = f.readline().strip()
+        line = f.readline().strip().decode("ascii", "ignore")
         if not line:
             break
         l = line.split(':', 1)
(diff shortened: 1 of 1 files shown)


kbandla:
looking..

dpkt/http.py
```diff
@@ -19,7 +19,7 @@ def parse_headers(f):
         # 2) If this is a weird header, which do not ends with '\r\n',
         #    f.readline() would return with '', then after strip(),
         #    we still get an empty string, also break the loop.
-        line = f.readline().strip()
+        line = f.readline().strip().decode("ascii", "ignore")
```
obormot: Giving the "ignore" to decode() is a common theme here, it doesn't feel "clean"..  @scott-hand can you explain the logic behind all the ignores?

brifordwylie: @obormot what would be the alternative? If you don't use ignore then anytime you have a decoding error an exception would be thrown, so are you suggestion this would be a flag? If so that sounds good/fine, I suggest the flag be defaulted to ignore, because at least IMO I believe that most use cases you'd like the parsing to continue even if a decode error is hit.

brifordwylie:
@kbandla can you qualify "(btw, some of this exact stuff was already done by @sunhao2014 . His notes here. )" ?

We've had an email thread on this class since Jan 8th and this is the first I heard that @sunhao2014 had already worked on this. I'm looking at his PR https://github.com/kbandla/dpkt/blob/migrate_py3/dpkt/http.py and I'm not seeing the equivalent functionality so perhaps I'm missing something.

dpkt/http.py
scott-hand: @obormot It looks like I'm doing it a lot but it's being done in only 3 cases:

1. Decoding to HTTP fields ("POST", "200", headers, etc.). Lines: 22, 164, 211
2. Encoding from HTTP fields. Lines: 127, 129, 188, 237
3. Convenience `str()` method. Lines: 120

This sentence from the RFC quote in my original PR explains my reasoning regarding doing it for the first two reasons:

>A recipient SHOULD treat other octets in field content (obs-text) as opaque data

You basically have a few options in Python 2 and 3. One is to ignore it, the other is to replace it with U+FFFD. I believe backslashreplace is available in 3 and in 2 (inside the codecs package), but I'm not sure juggling the codecs package vs. `decode()` is worth preserving "opaque" data.

Now, regarding the third reason, I don't have an RFC off the top of my head (and I'm not sure if there is one), but you have the option of inserting junk (replacement), omitting the character (ignore), or throwing an exception (strict).

I view `str()` as largely a convenience method, as users need to know how they are decoding data to do it correctly (looking for charset information for example). So the question is, should we have `str()` raise on decoding errors if it shouldn't be relied on for 100% correct information in the first place?

scott-hand:
@kbandla @brifordwylie 

Yes, just to be clear, I used that code as a starting point. His port will update the code so that it runs without compatibility errors (old style exceptions, `.iteritems()`, etc.), but the http.py module is still parsing raw data buffers as unicode strings. See my test case for an example of a common case that raises the issue.

brifordwylie:
just fwiw, I tested this branch with Chains for both python 2 and 3 and seems to work AOK.

```
HTTP_REQUEST
	Payload Size: 363
	192.168.1.14 --> Host: 192.168.1.8:2869 
	URI: /upnphost/udhisapi.dll?content=uuid:b8112991-56b1-4285-9dc5-9f21fe6c6ce5 
	Referer: - 
	Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.98 Safari/537.36

HTTP_RESPONSE
	Payload Size: 1341
	192.168.1.8 --> 192.168.1.14 
	URI: http://192.168.1.8:10247/apps/ 
	Headers: {'content-length': '1064', 'content-type': 'text/xml; charset="utf-8"', 'content-language': 'en-US', 'server': 'Microsoft-Windows/10.0 UPnP/1.0 UPnP-Device-Host/1.0 Microsoft-HTTPAPI/2.0', 'application-url': 'http://192.168.1.8:10247/apps/', 'date': 'Mon, 13 Mar 2017 23:45:02 GMT'}
```

dpkt/http.py
obormot: @scott-hand So it's possible I'm misinterpreting the RFC here, but in my understanding treating the data as opaque means leaving these characters as-is (?). Using 'ignore' means we'll be throwing them away. So then parsing a byte stream into HTTP instance and then applying str() to it will result in a different byte stream. I also suspect we lose ability to create malformed packets (e.g. intentionally insert non-ASCII chars inside headers). I need to think about it.

brifordwylie: @obormot I agree with your interpretation of the RFC, but thinking this through a bit.. I believe the following:
1) Implementing the functionality to leave the characters as-is will be interesting/complicated.
2) Another Issue/PR is probably the best route for that (IMO). I'm going to create a new issue now.

Currently the http module is broken for py3 and this PR puts us in a much better state.

scott-hand:
Ouch, looks like rerunning the test failed because Python 3 doesn't keep dict ordering.

ohboy

I'll update it with an OrderedDict now.

brifordwylie:
@scott-hand Python has never guaranteed iteration order of keys in a dict.. and they explicitly say that you should not rely on any order. So I agree that using an OrderedDict is probably good for http headers.... I just noting the fact that we we're probably 'getting away with something' in Python 2 :)

scott-hand:
@brifordwylie

This is true but anecdotally my experience is that they behave a little different in their arbitrariness for dicts so yeah OrderedDict is better. Sadly it doesn't exist in Python 2.6. Might pull in some sort of similar sort for that, or consider storing a list of tuples, but honestly I'd rather leave that for another PR since http.py has always used dicts.

kbandla:
tested with py3.4 & py2.7 against [http_google.pcap](http://chrissanders.org/captures/http_google.pcap) and [http.cap](https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=http.cap)
looks good 👍

brifordwylie:
Okay, so there's an open conversation about the handling of non-ascii http headers, I've created a ticket for that (https://github.com/kbandla/dpkt/issues/340). The current functionality will support/'do the right thing' for most use cases so I'm going to go ahead and merge this PR.

scott-hand:
@brifordwylie Thanks! This makes my Python 3 dpkt life much easier.

PR #338, opened 2017-03-14
add IP version and header length tests

dtrodrigues:


brifordwylie:
looks good.. thx @dtrodrigues

PR #339, opened 2017-03-14
add RTP tests

dtrodrigues:


dpkt/rtp.py
```diff
@@ -135,3 +135,36 @@ def unpack(self, buf):
...
+    rtp.version = 2
+    rtp.p = 0
+    rtp.x = 0
+    rtp.cc = 0
+    rtp.m = 0
+    rtp.pt = 3
+    rtp.seq = 1234
+    rtp.ts = 5678
+    rtp.ssrc = 0xabcdef01
+    assert (rtp.version == 2)
```
obormot: You're just assigning these values above, there's no encoding/decoding happening. So what's the point of these asserts? It would make sense to do e.g. 
```
rtp = RTP(bytes(rtp))
```
before all the asserts

dtrodrigues: Would it make sense to remove `rtp = RTP()` from line 152? This would mirror the setter tests from `test_frag()` in the ip.py file.

dtrodrigues: And also change the values from what they were in the byte string provided to the initial constructor to test that the values changed?

dtrodrigues: Or would it make more sense to call the setters, and then compare `bytes(rtp)` to the byte string it's supposed to have after those fields are changed?

obormot: Oh this tests custom setters, not just regular attributes. This actually makes sense now. I'd add a comment stating that this block tests setters. There's already a test for encoding and decoding above, so that's covered. I'd just add a comment in line 151 saying that the test below is for the setters and getters. Thanks!

obormot: @dtrodrigues I suggest an improvement here. Setting these fields to 0 doesn't really exercise the setters code. Something other than 0s would be better IMO

dtrodrigues: I'm updating it to set the `m` field to 1. However, changing any of the other fields changes the meaning of the rest of the packet. For instance, setting `p` to 1 would mean that the last byte of the payload tells how many bytes of padding were added to the end. Setting `cc` non-zero would mean that that many CSRCs follow the header.

Does it make sense to test the header setters independently of changing the payload data or the header extensions? In a larger sense, does it make sense to allow a user to, for instance, set `cc` to not be zero, but not enforce that the length of `csrc` is 4 times the value of `cc`? I'm not sure what the philosophy of dpkt is, but this would allow the user to change header fields and create an invalid RTP packet without warning. Is that okay?

PR #344, opened 2017-03-15
change a few dhcp header fields to be initialized with bytes

brifordwylie:
@obormot @kbandla need one approval and then I'll merge. Actually do we need to change some of the other classes as well? like ethernet.py for src and dst field initialization?


dpkt/dhcp.py
```diff
@@ -177,6 +177,11 @@ def test_dhcp():
     dhcp = DHCP(s)
     assert (s == bytes(dhcp))
 
+    # Test default construction
+    dhcp = DHCP()
+    assert isinstance(dhcp.chaddr, bytes)
```
obormot: how about when you initialize as ```DHCP(b'long strings of bytes')``` and then the same assert?

brifordwylie: sure... I'll just copy/paste those 3 asserts to the test above...  Also (question from above) do we need to do the same b'' initialization on things like src/dst in ethernet.py and others?

PR #347, opened 2017-03-17
cdp correction

mori-b:
CDP packet was seen as malformed in Wireshark, because of the Addresses field which was handled differently from the others. I don't know the reason for this special handling, but I made some changes and now Wireshark parses successfully the  CDP Addresses field.
I tested the changes and added a testing script under examples/
 

	modified:   dpkt/cdp.py
	new file:   examples/test_cdp.py
-correction cdp addresses field
-add test_cdp script under examples, to test cdp sending and
cdp pcap reading

### DIFF
--- a/dpkt/cdp.py
+++ b/dpkt/cdp.py
@@ -46,6 +46,7 @@ class CDP(dpkt.Packet):
     class Address(dpkt.Packet):
         # XXX - only handle NLPID/IP for now
         __hdr__ = (
+            ('numberadd', 'L', 1) , # number of address
             ('ptype', 'B', 1),  # protocol type (NLPID)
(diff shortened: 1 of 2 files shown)


brifordwylie:
@mori-b thanks for looking at this, a couple of things to mention:
1) I don't know jack about CDP so I have no idea if treating the address special should be done or not.
2) Tests right now (for better or worse) are just placed at the bottom of the file (see icmp.py for example). The examples directory is for tutorial like stuff and isn't actually part of the automated tests.

For 1) we'll have to get someone who knows about CDP.. @obormot @kbandla you guys know CDP?

For 2) please move the test when you get a chance :)

mori-b:
Thanks, I'll move the test asap

PR #348, opened 2017-03-18
full test coverage for aim module, more descriptive tlv parsing error

dtrodrigues:


dtrodrigues:
I'm not sure the best way to get the Python 3 error handling tests to work. In Python 2, it's `e.message` to get the error string but in 3 it's just `e` in `except FooError as e:`.

dtrodrigues:
Using `sys.version_info[0]` is an option, but seems rather kludgy.

dtrodrigues:
Looks like calling `str` on the error did the trick. These commits should probably be squashed while merging.

PR #352, opened 2017-03-23
CDP improvment

mori-b:
CDP parsing and construction improvement : 
- create well formed packets (previous version had malformed packets visible in Wireshark)
- generic TLVs are supported automatically, special TLVs support is easier to add

### DIFF
--- a/dpkt/cdp.py
+++ b/dpkt/cdp.py
@@ -7,6 +7,7 @@ import struct
 
 from . import dpkt
 
+
 CDP_DEVID = 1  # string
 CDP_ADDRESS = 2
 CDP_PORTID = 3  # string
--- a/dpkt/cdp.py
+++ b/dpkt/cdp.py
@@ -44,8 +44,6 @@ class CDP(dpkt.Packet):
         ('sum', 'H', 0)
     )
 
-    #keep here the TLV classes whose header is different from the generic TLV header (example : TLV_Addresses)
-    tlv_types = {CDP_ADDRESS: 'TLV_Addresses'}
 
     class TLV(dpkt.Packet):
         '''When constructing the packet, len is not mandatory : if not provided, then self.data must be this exact TLV payload'''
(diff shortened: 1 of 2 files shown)


mori-b:
Curiously the coverage test fails only in HTTP for python 3.5, even though the only changes are relative to CDP. Moreover this specific failing test ".tox/py35/bin/py.test --cov=dpkt dpkt" works on my local machine.

mori-b:
Second identical coverage test run passed ...

dpkt/cdp.py
```diff
@@ -43,69 +44,111 @@ class CDP(dpkt.Packet):
...
         dpkt.Packet.unpack(self, buf)
         buf = self.data
         l = []
         while buf:
-            tlv = self.TLV(buf)
-            l.append(tlv)
+            #find the right TLV according to Type value
+            tlv_find_type = self.TLV(buf).type
+            #if this TLV is not in tlv_types, use the default TLV class
+            tlv = getattr(self, self.tlv_types.get(tlv_find_type, 'TLV'))(buf)
```
obormot: Why getting class by name? It brings all this getattr() magic that seems unnecessary. You can just specify the actual class, e.g.
```
tlv_types = {CDP_ADDRESS: TLV_Addresses}
```

mori-b:
@obormot 
Answering your remark regarding the use of getattr : I removed it in last commit, the downside is that then I have to define a class attribute at the end of the class, so that it can access the nested classes

dpkt/cdp.py
mori-b: Done as you suggested in last commit. Only downside is that the tlv_types class attribute must now be defined at the end of the class body, in order to access the nested classes.

mori-b:
In spite of the fact I added tests to the CDP module, the automatic check shows a decrease of coverage. An explanation about its meaning could help, thanks.

obormot:
This .01% is likely (unfortunately) due to this block not getting executed during the test:
```
if __name__ == '__main__':
    test_cdp()
    print('Tests Successful...')
```

You can click on "Details" right of the coverals to see the detailed report from coverals.

brifordwylie:
@mori-b @obormot yeah.. lets not worry about the 0.01% coverage issue.. it's minutia... if the PR is good.. approve/merge it

mori-b:
@obormot I see, the coverage test runs directly the "test_cdp" function, without needing to invoke the "if __name__=='__main__'" code. It seems to be the right way, it's the same behavior as Icmp.
So it looks like the coverage decreased because of useful code most of which is rightfully not in the scope of tox.

PR #369, opened 2017-06-01
On windows, check if the file is opened in binary mode

kbandla:
Most window users open the files with the wrong mode. Unfortunately, the file mode property is read-only, so we can not fix it on the fly. Raise an exception instead.

### DIFF
--- a/dpkt/pcap.py
+++ b/dpkt/pcap.py
@@ -197,6 +197,8 @@ class Writer(object):
     """
 
     def __init__(self, fileobj, snaplen=1500, linktype=DLT_EN10MB, nano=False):
+        if fileobj.mode != 'wb':
+            raise ValueError('PCAP file not in binary mode (rb)')
         self.__f = fileobj
         self._precision = 9 if nano else 6
         magic = TCPDUMP_MAGIC_NANO if nano else TCPDUMP_MAGIC
@@ -240,6 +242,9 @@ class Reader(object):
 
     def __init__(self, fileobj):
         self.name = getattr(fileobj, 'name', '<%s>' % fileobj.__class__.__name__)
+        if sys.platform.startswith('win'):
+            if fileobj.mode != 'rb':
+                raise ValueError('PCAP file (%s) not opened in binary mode (rb)'%self.name)
         self.__f = fileobj
         buf = self.__f.read(FileHdr.__hdr_len__)
         self.__fh = FileHdr(buf)


brifordwylie:
Why are all the builds suddenly failing? Seems to be related to https://github.com/pypa/setuptools/issues/937. I'll poke around on it

dpkt/pcap.py
```diff
@@ -197,6 +197,8 @@ class Writer(object):
     """
 
     def __init__(self, fileobj, snaplen=1500, linktype=DLT_EN10MB, nano=False):
+        if fileobj.mode != 'wb':
```
obormot: add a check for Windows, same as below?

dpkt/pcap.py
```diff
@@ -240,6 +242,9 @@ class Reader(object):
 
     def __init__(self, fileobj):
         self.name = getattr(fileobj, 'name', '<%s>' % fileobj.__class__.__name__)
+        if sys.platform.startswith('win'):
+            if fileobj.mode != 'rb':
+                raise ValueError('PCAP file (%s) not opened in binary mode (rb)'%self.name)
```
obormot: pep8
```
... (rb)' % self.name)
         ^ ^ spaces
```

dpkt/pcap.py
```diff
@@ -240,6 +242,9 @@ class Reader(object):
 
     def __init__(self, fileobj):
         self.name = getattr(fileobj, 'name', '<%s>' % fileobj.__class__.__name__)
+        if sys.platform.startswith('win'):
+            if fileobj.mode != 'rb':
```
obormot: both mode checks feel a bit too restrictive (but make complete sense at the same time..).  I personally would go with smth like ```if 'b' not in fileobj.mode```

dpkt/pcap.py
kbandla: Oops! yes, thanks for catching

dpkt/pcap.py
kbandla: Just noticed that i did not account for append modes, your suggestion takes care of that too.

brifordwylie:
Okay the tests on TravisCI are now properly running again... there are two failed tests... 

```
    def __init__(self, fileobj, snaplen=1500, linktype=DLT_EN10MB, nano=False):
>       if fileobj.mode != 'wb':
E       AttributeError: 'BytesIO' object has no attribute 'mode'
```

and then a test for the deprecated decorator (I don't totally understand why that is failing).

kbandla:
going to look at this some more later today. I have a bunch of more updates for pcap.py anyway

brifordwylie (approved):
LGTM :)

brifordwylie:
Shrug... somehow the builds are failing... will have to circle back to this and see if I can figure out the issues...

PR #370, opened 2017-06-01
see if an upgrade to setuptool helps travis not fail

brifordwylie:


obormot:
I think this is outdated.. Travis works fine; closing.

Issue #373, opened 2017-06-09
Fix Deprecation test


brifordwylie:
Newest version of pytest (3.1.2) does something 'different' with stdout/err than (3.0.7)
-See PR #374 

```
>           assert ('DeprecationWarning: Call to deprecated method deprecated_decorator' in out.getvalue())
E           AssertionError: assert 'DeprecationWarning: Call to deprecated method deprecated_decorator' in ''
E            +  where '' = <built-in method getvalue of cStringIO.StringO object at 0x27f7998>()
E            +    where <built-in method getvalue of cStringIO.StringO object at 0x27f7998> = <cStringIO.StringO object at 0x27f7998>.getvalue
```

crocogorical:
All deprecated methods removed in #354. Can this be closed?

brifordwylie:
👍

PR #374, opened 2017-06-09
Newest version of pytest (3.1.2) does something 'different' with stdout/err than (3.0.7)

brifordwylie:


brifordwylie:
@kbandla @obormot so I thought this was one thing and it turned out to be another.

Between the 3.0.7 version of pytest and the 3.1.2 version of pytest something has changed about how stdout/err is handled so that the tests ALL fail under ANY version of Python. The out.getvalue() returns the empty string when run under the newer version of pytest (works fine when run standalone or with older version of pytest).

If you guys know anything, please chime in.. probably won't have time to dive in right away.

brifordwylie:
I'm going to merge this PR because temporarily hacking around the deprecation test is better than having failing build. I'll keep the Issue #373 open and I'll circle back around to it.

PR #378, opened 2017-07-27
Use ciphersuite objects instead of int list

c0r0n3r:


obormot:
this change is included in another (larger) PR from the same author, see #384.
closing as duplicate

PR #382, opened 2017-08-01
Adding flush method for dpkt.pcap.Writer class

rootkiter:
With the flush function, the contents of the buffer can be written to the file at any time to avoid data loss...


rootkiter:
why not successful... I didn't see any wrong...

obormot:
> why not successful... I didn't see any wrong...
You added a method without adding a unit test for it. So coverage has decreased and Coveralls isn't happy.

On another note, do you mind adding the same to pcapng.py to keep the interfaces consistent?

obormot:
Closing due to age and lack of follow-up. Same can be achieved by calling `.flush()` on the file handle outside of the pcap.Writer class

Issue #389, opened 2017-09-22
We're not doing 802.1q and 802.1ad correctly in ethernet.py


smutt:
Currently in ethernet.py if a packet has an 802.1q header we replace the outer ethertype with the one from the 802.1q header. This happens at ethernet.py line 89. This should not happen.

Also, 802.1ad (commonly called QinQ, or Virtual Metro Area Networks(VMAN)) is not done by stacking 0x8100 ethertypes. Cisco recommends 0x9100 or 0x9200 in its QinQ docs[1]. I think it is rather weak on Cisco's part that they recommend this, as they should be adhering to 802.1ad, but whatever. The 802.1ad standard says to use 0x88a8.[2] It can theoretically be anything, which is of course a problem, but I think we can do better. 

The code in ethernet.py lines 85-93 takes none of this into account and instead assumes a stacking of ethertypes 0x8100, which I doubt anyone is actually doing. Because it would mean any intermediate switch wouldn't know if a frame is single or double tagged.

I'm fine developing the code for all this, but before I propose any PR I want to get the opinion of the devs on this. Because I suspect there is a reason for the current behavior.

I'm assuming for now that dpkt will never support 802.1ah, which would be a serious pain. Luckily I think it's mostly been obviated by Nortel's death and the rise of VPLS.

[1] https://www.cisco.com/c/en/us/td/docs/routers/10000/10008/feature/guides/qinq.html
[2] https://en.wikipedia.org/wiki/IEEE_802.1ad

smutt:
I just saw that ethernet.py does a similar thing on line 107 for MPLS.
 self.type = ETH_TYPE_IP

You can't just overwrite the outside ethertype with the inner ethertype. It clobbers that data and then it's impossible write the original packet back to disk.

I'll need to investigate everywhere a statement like self.type = ETH_TYPE_X is made in ethernet.py. There's another one in the IPX decapsulation, but I don't really know enough about IPX to understand the implications.

PR #390, opened 2017-09-23
Fixes proposed for issue #389

smutt:
These are my proposed fixes for issue #389. I'm curious what obormot thinks of them since he/she implemented PR #267. I'm also very willing to have a discussion about this and take a different approach if necessary.

### DIFF
--- a/dpkt/ethernet.py
+++ b/dpkt/ethernet.py
@@ -49,7 +49,9 @@ ETH_TYPE_PPPoE_DISC = 0x8863  # PPP Over Ethernet Discovery Stage
 ETH_TYPE_PPPoE = 0x8864  # PPP Over Ethernet Session Stage
 ETH_TYPE_LLDP = 0x88CC  # Link Layer Discovery Protocol
 ETH_TYPE_TEB = 0x6558  # Transparent Ethernet Bridging
-
+ETH_TYPE_8021ad = 0x88a8 # 802.1ad
+ETH_TYPE_QINQ1 = 0x9100 # Legacy QinQ
+ETH_TYPE_QINQ2 = 0x9200 # Legacy QinQ
 
 class Ethernet(dpkt.Packet):
     """Ethernet.
@@ -78,18 +80,30 @@ class Ethernet(dpkt.Packet):
                 self._unpack_data(self.data)
 
     def _unpack_data(self, buf):
-        if self.type == ETH_TYPE_8021Q:
-            self.vlan_tags = []
+        self.inner_type = None
+        if self.type == ETH_TYPE_8021ad or self.type == ETH_TYPE_QINQ1 \
+            or self.type == ETH_TYPE_QINQ2:
 
-            # support up to 2 tags (double tagging aka QinQ)
+            # 802.1ad and QinQ support up to 2 tags
             for _ in range(2):
                 tag = VLANtag8021Q(buf)
                 buf = buf[tag.__hdr_len__:]
                 self.vlan_tags.append(tag)
-                self.type = tag.type
-                if self.type != ETH_TYPE_8021Q:
+                self.inner_type = tag.type
+                if tag.type != ETH_TYPE_8021Q:
                     break
+
             # backward compatibility, use the 1st tag
+            self.vlanid, self.priority, self.cfi = self.vlan_tags[0].as_tuple()
+
+        elif self.type == ETH_TYPE_8021Q:
+            self.vlan_tags = []
+            tag = VLANtag8021Q(buf)
+            buf = buf[tag.__hdr_len__:]
+            self.vlan_tags.append(tag)
+            self.inner_type = tag.type
+
+            # backward compatibility
             self.vlanid, self.priority, self.cfi = self.vlan_tags[0].as_tuple()
 
         elif self.type == ETH_TYPE_MPLS or self.type == ETH_TYPE_MPLS_MCAST:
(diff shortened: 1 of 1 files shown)


smutt:
test_eth_802dot1q() is failing because pack() needs to be fixed as well. I'll work on this.

test_eth_802dot1q_stacked() is an invalid test case. Both 802.1q headers have 0x8100 as their ethertype. I'll rewrite this test.

test_eth_mpls_stacked() I'm not really sure about. I'm not familiar enough with how this is supposed to work. I'll look into it.

obormot:
After looking at how QinQ decoding happens today it's clear we could do better. I have not tested this PR yet but I'd like to address some of the unit test changes. Specifically:

> test_eth_802dot1q_stacked() is an invalid test case. Both 802.1q headers have 0x8100 as their ethertype. I'll rewrite this test.

... not invalid;  first 0x8100 type is in the Ethernet header;  the 2nd 802.1q header has IP.  the bytes were copied from a real pcap originally found [http://packetlife.net/captures/802.1Q_tunneling.cap](here)

I've extracted a single packet used in the test 
[https://drive.google.com/open?id=12cvGfkZszTaWaFDXa_lMGrDuc-ewgoo7](here)

So we need to make sure the test passes without modifying the input data

obormot:
I've checked a few more test cases with double 802.1Q tags, and MPLS tags, and it looks like with the proposed changes only one tag is getting decoded:

```
# this contains 2 802.1Q tags, only one gets decoded

In [11]: %paste
p1 = (
    b'\xff\xff\xff\xff\xff\xff\xca\x03\x0d\xb4\x00\x1c\x81\x00\x00\x64\x81\x00\x00\xc8\x08\x06'
    b'\x00\x01\x08\x00\x06\x04\x00\x01\xca\x03\x0d\xb4\x00\x1c\xc0\xa8\x02\xc8\x00\x00\x00\x00'
    b'\x00\x00\xc0\xa8\x02\xfe\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
)

In [12]: Ethernet(p1)
Out[12]: Ethernet(dst='\xff\xff\xff\xff\xff\xff', src='\xca\x03\r\xb4\x00\x1c', type=33024, 
vlan_tags=[VLANtag8021Q(type=2054, pri=0, id=200, cfi=0)], 
data=ARP(sha='\xca\x03\r\xb4\x00\x1c', spa='\xc0\xa8\x02\xc8', tha='\x00\x00\x00\x00\x00\x00', tpa='\xc0\xa8\x02\xfe', data='...'))

# supposed to be:
vlan_tags=[
        VLANtag8021Q(type=33024, pri=0, id=100, cfi=0), 
        VLANtag8021Q(type=2054, pri=0, id=200, cfi=0)
    ],

# the 1st tag is missing
```

dpkt/ethernet.py
```diff
@@ -235,6 +231,42 @@ def get_type(cls, t):
     def get_type_rev(cls, k):
         return cls._typesw_rev[k]
 
+    # Property functions needed for backwards compatibility
+    def get_vlanid(self):
+        return self.vlan_tags[0].id
```
obormot: (applies to all the new properties here) indexing without checking if self.vlan_tags is empty

smutt:
In 802.1ad the outer tag(first tag/S-tag) will typically have an ethertype of 0x88a8. That's what the IEEE standard suggests. But it needs to be something other than 0x8100 at the very least.

Wikipedia gets this right.
https://en.wikipedia.org/wiki/IEEE_802.1ad

Maybe Cisco did something different for a while in QinQ, before 802.1ad became standardized. I'm more familiar with Extreme Networks gear which used 0x9100 as default before 802.1ad was standardized, then switched over to using 0x88a8 as default.

If the ethertype of the outer tag is 0x8100 then the correct thing to do is to only decode that tag. In fact, one might even say it is a broken packet because how could a switch know that it needs to be treated differently than a normal single tagged packet? A receiving switch will need to act on the tags based on their order and ethertypes.

obormot:
@smutt so to recap the outer tag could be 0x88a8 or 0x9100 or 0x9200 then this is double tagging and the inner tag is always 0x8100. I have found a pcap with 802.1ad frame but none of the 0x9X00 frames. 
Alternatively both tags are 0x8100 (Wikipedia mentions it as "C-tag stacking on C-tag"; legacy, but it's possible to have packets like these). I've found a couple pcaps with this, so I think dpkt should decode them both. I don't think dpkt should care about the switches' behavior (prob. different anyway, depending on the vendor), its job is to decode the packet in layers according to the structure. 
I wasn't sure if you were still interested in working on this and have spent some time to address the issue in my own PR #426 ("inspired" by this PR and the discussions.. thank you very much!). I appreciate if you take a look; alternatively we could keep iterating on this PR (fix/add unit tests etc.)

smutt:
Hi Oscar,

Thanks for doing all this work and I think your approach makes sense for dpkt. I was maybe getting hung up on the fact that the IEEE 802.1ad standard says it MUST be 0x88a8. But I agree you're right in that dpkt SHOULD be ethertype agnostic. 

The project I was working on that lead me to suggest these changes for dpkt has somewhat fallen off my radar ATM. And to be honest I need to refamiliarize myself with the code I wrote for this PR before I can comment on your changes. I should have some time this week to look at this and get back to you. Please let me spend a bit more time on reviewing this before merging, and I promise to get back to you this month(November). Thanks again for spending your time on this.

smutt:
Hi Oscar,

I like your changes. Please reject this PR and merge #426.

Thanks again for taking the time to work on this.

--Andrew

Issue #391, opened 2017-10-28
Inefficient packet writing


crocogorical:
Each packet write checks the system endianness, and constructs an LE/PktHdr, to allow the packet to be written. 

I am not aware of conditions under which system endianness could change between packet writes, so this testing should be moved to the Writer initialization instead.

crocogorical:
PR merged. Closing.

PR #392, opened 2017-10-28
Speed up packet writing for pcap. Add writepkts method.

crocogorical:
Speed up the packet output process and add additional methods to improve performance under certain conditions to address #391 

This works by moving endianness testing to Writer initialization and creating a partial function called `_pack_hdr` on the `Packet` class which allows much faster access to the struct packing method with an already constructed `__hdr_fmt__`. 

The original writepkt method is available, with the same signature, but improved performance. In addition two new methods are added.

writepkt_time(self, pkt, ts)
--------------------------------
By taking a non-optional timestamp, we can skip the test for timestamp and gain a marginal speed improvement. As this is a new method, the bytes() call in writepkt is omitted, gaining another marginal speed improvement.

writepkts(self, pkts)
-------------------------
pkts must be an iterable of tuples (ts, pkt). Although this is the other way round from writepkt, and therefore possibly confusing, it is the same order as readpkts. This means that the mass input and output functions are of the same format.

testing
======

The motivation for this work is speed. Writing 1 million packets with timestamps gives an average of 80% time reduction (500% speed increase).

Test was performed using a generator to produce the same packet, and `time.perf_counter()` to give the time measurements.

Method at [this gist](https://gist.github.com/crocogorical/986ca4e2fe82dc834fbdad498ce6ecf4), with `80c8545f6537ecd56b3ce50c8a6984291a73b666` imported as `dpkt_master` for comparison.

```
TIMES 5, PACKETS 1000000
perf   perf%   stdev  doc
28.629   0.00%  1.008 original writepkt(pkt, ts)
28.265  -1.27%  0.187 original writepkt(pkt)
 8.594 -69.98%  0.026 writepkt(pkt, ts)
10.176 -64.46%  0.052 writepkt(pkt)
 7.488 -73.85%  0.020 writepkt_time(pkt, ts)
 7.074 -75.29%  0.253 writepkts([(ts, pkt)])
 5.605 -80.42%  0.027 writepkts(buf_list)
```

### DIFF
--- a/dpkt/pcap.py
+++ b/dpkt/pcap.py
@@ -199,30 +199,47 @@ class Writer(object):
     def __init__(self, fileobj, snaplen=1500, linktype=DLT_EN10MB, nano=False):
         self.__f = fileobj
         self._precision = 9 if nano else 6
+        self._precision_multiplier = 10**self._precision
+
         magic = TCPDUMP_MAGIC_NANO if nano else TCPDUMP_MAGIC
         if sys.byteorder == 'little':
             fh = LEFileHdr(snaplen=snaplen, linktype=linktype, magic=magic)
+            self._PktHdr = LEPktHdr()
         else:
             fh = FileHdr(snaplen=snaplen, linktype=linktype, magic=magic)
+            self._PktHdr = PktHdr()
+
+        self._pack_hdr = self._PktHdr._pack_hdr
         self.__f.write(bytes(fh))
 
     def writepkt(self, pkt, ts=None):
+        """ Write pkt and optional ts to file """
         if ts is None:
             ts = time.time()
         s = bytes(pkt)
-        n = len(s)
-        sec = int(ts)
-        usec = int(round(ts % 1 * 10 ** self._precision))
-        if sys.byteorder == 'little':
-            ph = LEPktHdr(tv_sec=sec,
-                          tv_usec=usec,
-                          caplen=n, len=n)
-        else:
-            ph = PktHdr(tv_sec=sec,
-                        tv_usec=usec,
-                        caplen=n, len=n)
-        self.__f.write(bytes(ph))
-        self.__f.write(s)
+        self.writepkt_time(s, ts)
+
+    def writepkt_time(self, pkt, ts):
+        """ Write pkt and mandatory ts to file """
+        self.writepkts([(ts, pkt)])
+
+    def writepkts(self, pkts):
+        """
+        Take an iterable of (ts, pkt), and write to file.
+        """
+        fd = self.__f
+        pack_hdr = self._pack_hdr
+        precision_multiplier = self._precision_multiplier
+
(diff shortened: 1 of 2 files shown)


crocogorical:
There is now 100% test coverage of the pcap.Writer class.

dpkt/pcapng.py
```diff
@@ -283,13 +285,29 @@ def unpack(self, buf):
...
+                n,
+                self.iface_id,
+                self.ts_high,
+                self.ts_low,
+                pkt_len,
+                pkt_len,
+                n
+            )
+
+        return b''.join([hdr_buf[:-4], _padded(pkt_buf), opts_buf, hdr_buf[-4:]])
```
obormot: should ```_PcapngBlock.__bytes__``` be changed in similar fashion?

dpkt/pcapng.py
```diff
@@ -368,15 +388,66 @@ def writepkt(self, pkt, ts=None):
...
+            pkt_len=n,
+            pkt_data=s
+        )
         self.__f.write(bytes(epb))
 
+    def writepkts(self, pkts):
+        """
+        Take an iterable of (ts, pkt), and write to file.
+        """
+
```
obormot: redundant whitespace

crocogorical:
Thanks for the review @obormot ! No disagreements, I'll hopefully submit patches tomorrow.

crocogorical:
CI build will pass when #404 is resolved.

dpkt/pcapng.py
```diff
@@ -154,10 +154,17 @@ def _do_pack_options(self):
...
-        return hdr_buf[:-4] + opts_buf + hdr_buf[-4:]
+        n = len(opts_buf) + self.__hdr_len__
+        self.len = n
+        self._len = n
+
+        hdr_buf = self._pack_hdr(
+            self.type,
+            n,
+            n,
+        )
```
obormot: I personally think it's a waste of space. It could fit in 1 line

dpkt/pcap.py
```diff
@@ -216,19 +216,38 @@ def __init__(self, fileobj, snaplen=1500, linktype=DLT_EN10MB, nano=False):
...
+
+        Args:
+            pkt: Some `bytes` to write to the file
+            ts: Timestamp in seconds. Defaults to current time.
+        """
         if ts is None:
             ts = time.time()
         s = bytes(pkt)
-        self.writepkt_time(s, ts)
+        self.writepkt_time(pkt, ts)
```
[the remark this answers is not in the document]
crocogorical: I will put the cast to bytes back in, as that is the original implementation. In the other methods (`writepkt_time` and `writepkts`) there is no cast as these are new methods and won't break any existing code.

Perhaps the call to `bytes` could be removed longer term, although this is a breaking change, and if speed is of concern, the new methods are also available.

Issue #375, opened 2017-06-18
Setting .text for a DNS TXT object throws exception in Python3


jjo-sec:
>   File "/usr/local/lib/python3.5/dist-packages/dpkt/dns.py", line 375, in __len__
>     return len(str(self))
>   File "/usr/local/lib/python3.5/dist-packages/dpkt/dpkt.py", line 141, in __str__
>     return str(self.__bytes__())
>   File "/usr/local/lib/python3.5/dist-packages/dpkt/dns.py", line 386, in __bytes__
>     buf = self.pack_rr(buf, rr)
>   File "/usr/local/lib/python3.5/dist-packages/dpkt/dns.py", line 342, in pack_rr
>     rdata = rr.pack_rdata(len(buf) + len(name) + 10, self.label_ptrs)
>   File "/usr/local/lib/python3.5/dist-packages/dpkt/dns.py", line 280, in pack_rdata
>     for x in self.text])
> TypeError: sequence item 0: expected a bytes-like object, str found

Using a byte array instead of a python3 string when setting .text generates a different exception

> TypeError: object of type 'int' has no len()

Was not able to determine a proper fix for the issue that makes both python2 and 3 happy without needing to check the version_info to perform the operation in a different manner depending on python version

brifordwylie:
Thanks @arbor-jjones we'll try to peek at this soon... can you put a small script that reproduces the error into this Issue?

crocogorical:
This will raise a `TypeError: object of type 'int' has no len()`
```python3
DNS.RR(type=DNS_TXT, text=b'Noonien Soong').pack_rdata(0, {})
```

This will raise a `TypeError: sequence item 0: expected a bytes-like object, str found`
```python3
DNS.RR(type=DNS_TXT, text='Noonien Soong').pack_rdata(0, {})
```

obormot:
Looks like this was addressed, closing. `.text` needs to be a list:
```
def test_rdata_TXT():
    rr = DNS.RR(
        type=DNS_TXT,
        text=[b'v=spf1 ptr ?all', b'a=something']
    )
    packdata = rr.pack_rdata(0, {})
    correct = b'\x0fv=spf1 ptr ?all\x0ba=something'
    assert packdata == correct
```
Test by the OP works as well, when changed to a list:
```
>>> DNS.RR(type=dpkt.dns.DNS_TXT, text=[b'Noonien Soong']).pack_rdata(0, {})
b'\rNoonien Soong'
```

PR #395, opened 2017-11-11
Pack rdata for DNS_TXT and DNS_HINFO correctly in python3. #375

crocogorical:
Use `struct.pack('B', len(x))` instead of `chr(len(x))` to create the length field in the packed TXT/HINFO records. Works the same across python2/3.

### DIFF
--- a/dpkt/dns.py
+++ b/dpkt/dns.py
@@ -276,8 +276,12 @@ class DNS(dpkt.Packet):
                 return struct.pack('>H', self.preference) + \
                        pack_name(self.mxname, off + 2, label_ptrs)
             elif self.type == DNS_TXT or self.type == DNS_HINFO:
-                return b''.join(['%s%s' % (chr(len(x)), x)
-                                for x in self.text])
+                return b''.join(
+                    b'%s%s' % (
+                        struct.pack('B', len(element)), 
+                        element,
+                    ) for element in self.text
+                )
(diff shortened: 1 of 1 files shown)


dpkt/dns.py
```diff
@@ -276,8 +276,10 @@ def pack_rdata(self, off, label_ptrs):
                 return struct.pack('>H', self.preference) + \
                        pack_name(self.mxname, off + 2, label_ptrs)
             elif self.type == DNS_TXT or self.type == DNS_HINFO:
-                return b''.join(['%s%s' % (chr(len(x)), x)
-                                for x in self.text])
+                return b''.join(
+                    struct.pack('B', len(element)) + element
```
obormot: (minor) personally I think 'element' is not a great var name here. it's just a symbol / char, I'd rather leave it as an 'x'

dpkt/dns.py
```diff
@@ -485,6 +487,28 @@ def test_txt_response():
     assert str(my_dns) == str(buf)
     assert bytes(my_dns) == buf
 
+def test_rdata_TXT():
+    rr = DNS.RR(
+        type=DNS_TXT,
+        text=[b'v=spf1 ptr ?all', b'a=something']
+    )
+
```
obormot: lines 495, 498, 506, 510 - extra whitespace, worse readability IMO. Other tests are formatted differently

Issue #396, opened 2017-11-11
Tox setup should match Travis CI build


crocogorical:
If locally running tests pass, the CI build should too. Not running python3.4 locally means that this sometimes doesn't happen.

Tox config should include `py34`.

obormot:
#397 is merged, closing

PR #397, opened 2017-11-11
Local tests should run against python3.4, as the travis CI build does.

crocogorical:
#396


brifordwylie:
A bit late on this... but perhaps @obormot and @kbandla should think about deprecating 3.4 and adding 3.6.. (that's what I'd vote for). If you decide not to deprecate 3.4 (which you should :) then at a minimum add 3.6 to both local and CI testing...

obormot:
Thanks @brifordwylie. Created #410 to follow up.

PR #403, opened 2017-11-25
update ip6.py to read all extension headers

mimta:
with current ip6.py, dpkt will escape all extension headers in packet and only read the last extension header in each packet. that mean with current version you are not able see extension headers in a packet except the last one.

**at the moment you only get this**

> <bound method IP6.unpack of IP6(v=6, fc=0, flow=8, plen=1182, nxt=60, hlim=255, src='Src', dst='Dst', p=58, extension_hdrs={60: IP6DstOptsHeader(nxt=58, length=8, options=[{'opt_length': 0, 'data': '', 'type': 31}], data='DATA')}, data=ICMP6(type=128, sum=754, data=Echo(seq=8, data='DATA')))>

**with new code you can read all extension headers**

screen shot 2017-11-25 at 20 13 38

with this update you are able to count/read all extension headers.

hope this help :)

### DIFF
--- a/dpkt/ip6.py
+++ b/dpkt/ip6.py
@@ -58,6 +58,7 @@ class IP6(dpkt.Packet):
     def unpack(self, buf):
         dpkt.Packet.unpack(self, buf)
         self.extension_hdrs = {}
+        self.all_extension_headers=[]
 
         if self.plen:
             buf = self.data[:self.plen]
@@ -69,6 +70,7 @@ class IP6(dpkt.Packet):
         while next_ext_hdr in ext_hdrs:
             ext = ext_hdrs_cls[next_ext_hdr](buf)
             self.extension_hdrs[next_ext_hdr] = ext
+            self.all_extension_headers.append(ext)
             buf = buf[ext.length:]
             next_ext_hdr = getattr(ext, 'nxt', None)
 


obormot:
@mimta I see the current ```extension_hdrs``` is a dict suggesting that it will save only the last extension header if there are multiple of the same type. Could you please attach a sample pcap / buffer, or better add a unit test covering the change?

mimta:
Hi
I've attached 2 pcap. 1 with 128 extension headers (all of them same type) and the other one with 7 extension headers (different types).
however with current code for both pcap, I could only get the last extension headers.

here is the simple code that would count the total number of extension here in a packet and separate them by type:

```
def packet_extension_headers_count(ip_packet):
    return len(ip_packet.all_extension_headers)


def print_packet_plus_extension_headers(ip_packet):
    count_of_packet_types={}

    for packet in ip_packet.all_extension_headers:
        key = __builtin__.type(packet).__name__
        if key in count_of_packet_types:
            count_of_packet_types[key]+=1
        else:
            count_of_packet_types[key]=1
    for key in count_of_packet_types:
        print key, count_of_packet_types[key]

print 'Total Number Of Extension Headers: ', packet_extension_headers_count(ip_packet)
print_packet_plus_extension_headers(ip_packet)
```

[github.zip](https://github.com/kbandla/dpkt/files/1510082/github.zip)

obormot:
merged in #412, closing

PR #416, opened 2018-06-01
BGP: Support for EVPN packet types 1-4

ngrauel:
I've added support for parsing BGP-EVPN packet types 1-4.  Information on EVPN packets can be found here: https://tools.ietf.org/html/rfc7432


dpkt/bgp.py
```diff
@@ -670,17 +676,87 @@ def __bytes__(self):
...
+        if self.type != 0x3:
+            self.esi = tmp[:10]
+            tmp = tmp[10:]
+
+        if self.type != 0x4:
+            self.eth_id = tmp[:4]
+            tmp = tmp[4:]
+
+        if self.type == 0x2:
+            self.mac_address_length = ord(tmp[0])
```
obormot: here and line 713: use compat_ord instead (from dpkt.compat)

dpkt/bgp.py
```diff
@@ -670,17 +676,87 @@ def __bytes__(self):
...
 
+class RouteEVPN(dpkt.Packet):
+    __hdr__ = (
+        ('type', 'B', 0),
+        ('len', 'B', 0)
+    )
+
+    def unpack(self, buf):
+        dpkt.Packet.unpack(self, buf)
+        self.data = self.data[:self.len]
```
obormot: `self.data` is normally the leftovers of what's left after the `unpack()` has finished working; 
I'd also suggest advancing `buf` directly, instead of `tmp` just for readability but that's minor. 
something like 
```
def unpack(self, buf):
     dpkt.Packet.unpack(self, buf)
     buf = self.data[:self.len]
     self.data = self.data[self.len:]
     # ... do the unpacking, advance buf as  buf = buf[8:]  etc.  
     # assuming the unpacking consumes self.len bytes entirely
```

dpkt/bgp.py
ngrauel: @obormot I changed my code to convert the buffer to a bytearray to make it behave the same in both python2/python3 before I saw this.  I then removed the calls to ord() completely.  Is compat_ord() a better way of doing this?

ngrauel: @obormot After looking at it again compat_ord() is definitely the cleaner way to do this so I'm updating my code.

dpkt/bgp.py
ngrauel: @obormot I updated self.data to only store the leftover bytes and added a self.route_data attribute to store the bytes that make up the evpn header so they can be easily returned without having to rebuild the header bytes.  I also changed tmp to buf.

dpkt/bgp.py
```diff
@@ -9,6 +9,7 @@
 
 from . import dpkt
 from .decorators import deprecated
+from dpkt.compat import compat_ord
```
obormot: `from .compat`

dpkt/bgp.py
```diff
@@ -670,17 +677,87 @@ def __bytes__(self):
...
+            self.eth_id = buf[:4]
+            buf = buf[4:]
+
+        if self.type == 0x2:
+            self.mac_address_length = compat_ord(buf[0])
+            if self.mac_address_length == 48:
+                self.mac_address = buf[1:7]
+                buf = buf[7:]
+            else:
+                self.mac_address = None
```
obormot: so in this case `self.mac_address` is None, but `self.mac_address_length` still holds a value? 
is `mac_address_length` needed as member of the class?  could it be just a local variable?
same goes for `ip_address_length` below

ngrauel: Yes, the mac_address_length field is always set to either 0 or 48.  If set to 48 then there is a mac_address field following it.  If set to 0 then there is no mac_address field so I set the value to None.  The ip_address_length/ip_address fields work the same way for the EVPN packet types they are included in.  I set them as class attributes so that those fields would be easily accessible in the parsed RouteEVPN object.

Issue #419, opened 2018-09-20
Radiotap module does not correctly decode radiotap format


kristahi:
There are multiple issues with the radiotap header decoder module, which make it more or less unusable at the moment for anything but getting at the 802.11 frames trailing the radiotap header.

Major issues:
- *Endianness is reversed*
Radiotap is defined as being little-endian, yet all multibyte fields are unpacked as big-endian. Obviously this breaks all such fields.
Importantly, this also applies to the bit ordering in the field presence bitmap.
There is a workaround of sorts already in place for decoding encapsulated data (`ntohs` is called on the raw value of the length field in `Radiotap.unpack`).
See also #132.
- *Parser only handles a single present fields bitmap*
Radiotap supports the inclusion of an arbitrary number of fields by "chaining" the presence bitmap when the most significant bit is set. This also allows fields of the same type to be repeated (e.g. to specify antenna data on radios that have multiple antennas). It isn't possible to reliably decode a radiotap header without handling this properly.
- *The tests pass because the test data are invalid radiotap headers*
Other radiotap parsers (Wireshark etc.) choke on the raw buffers supplied in the tests, which are big-endian encoded.

I am willing to provide patches to sort this out, but the extent of the breakage presents some challenges:
- If anyone is actually using the module somehow (besides to just decode 802.11 frames), they are surely working around the problems described above themselves, so by actually returning correct output it is possible existing programs will break.
- Also, writing a proper parser is not entirely trivial and there is already an existing, ISC-licensed Python parser library: https://github.com/radiotap/python-radiotap/
I'm unsure if it is worth the effort the duplicate this work. dpkt currently has no dependencies and I suppose this is deliberate?

One possibility is to just remove support for parsing the contents of the radiotap header and instead offer it in raw form so users can pass it into another library such as the one above themselves. This way you can still use dpkt to decode the actual frames.

Any opinions on this?

brifordwylie:
@kristahi as you may have noticed we're working with a bit of a skeleton crew here. I'll give you my input FWIW. 

Looking at https://github.com/radiotap/python-radiotap there are basically just two files, one approach is to contact the author, let him know we're interested in his work and would like to include them into dpkt and add him as a contributor :) @bcopeland (https://github.com/bcopeland)

Since I don't know anything about radiotap, you'd have to manage pulling in that work (if Bob would like to contribute it).

brifordwylie:
[...]

PR #423, opened 2018-11-09
Implement decoding of IEEE 802.11 DELBA Mgmt Action (Delete Block Ack).

kristahi:
Add support for decoding IEEE 802.11 DELBA (Delete Block Ack) management action frames.

There was only support for the block ack request and response actions, which caused an exception on captures that contained DELBA frames.

The Cisco equipment I saw generating these in the wild didn't quite follow 802.11-2016
when it came to the frame format, so I implemented what worked in practice. Seems to be in line with the Wireshark 802.11 dissector.


kristahi:
I was adding a test, which failed, which led to me discovering there are a bunch of endianness issues in the IEEE80211 module, similarly to the issues with Radiotap (see #419 ).

Funnily enough some of these are also present in the way Wireshark displays some IEEE 802.11 fields (e.g. the Frame Control field, but not the type/subtype subfields which are actually part of the FC).

Since there is a bunch of logic which is based on interpreting things with reversed endianness it is a little bit of work to fix it. I don't have time right now, but I will try to revisit when I have some more cycles available...

obormot:
@kristahi changes look good but the build fails in one of the unit tests.. please address before we could merge

obormot:
I think this fails because the unit test might have an error in it
```
assert ieee.action.block_ack_delba.delba_param_set == 0x0800
```
since the protocol is little endian and we're seeing `0800` in network order in the buffer, it should be comparing to an `8`, not `0800`.

Another minor thing I noticed is DELBA seems to stand for "Delayed block ack" not "Delete block ack".

obormot:
Found a sample, that's how Wireshark sees it.  `0008` on the wire, `0x0800` decoded value of the parameter set.
This corresponds with what's happening in the unit test.  So the unit test appears correct.  Means the DELBA decoder isn't - needs to be fixed.

image

obormot:
Superseded by https://github.com/kbandla/dpkt/pull/516, closing

Issue #442, opened 2019-01-10
Mostly zero ethernet frame throws exception


kylekeppler:
The attached pcap has a single packet that causes a `NeedData` exception in dpkt 1.9.2.

[capture.zip](https://github.com/kbandla/dpkt/files/2747090/capture.zip)

The frame looks like:

```
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | 16
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | 32
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 | 48
00 00 00 00 00 00 00 00 00 00 00 00 08 89 12 04 | 64
```

Example:

```python
for timestamp, buf in dpkt.pcap.Reader(f):
    eth = dpkt.ethernet.Ethernet(buf)
    print(eth)

```

Since the `ethertype` field is 0, it is being treated as logical link control, and throwing an exception:

```python
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
c:\workspace\dpkt\dpkt\dpkt\dpkt.py in __init__(self, *args, **kwargs)
     88             try:
---> 89                 self.unpack(args[0])
     90             except struct.error:

c:\workspace\dpkt\dpkt\dpkt\llc.py in unpack(self, buf)
     35
---> 36         dpkt.Packet.unpack(self, buf)
     37         if self.is_snap:

c:\workspace\dpkt\dpkt\dpkt\dpkt.py in unpack(self, buf)
    170         for k, v in compat_izip(self.__hdr_fields__,
--> 171                                 struct.unpack(self.__hdr_fmt__, buf[:self.__hdr_len__])):
    172             setattr(self, k, v)

error: unpack requires a buffer of 3 bytes

During handling of the above exception, another exception occurred:

NeedData                                  Traceback (most recent call last)
<ipython-input-5-a483b3f3f387> in <module>
      1 for timestamp, buf in dpkt.pcap.Reader(f):
----> 2     eth = dpkt.ethernet.Ethernet(buf)
      3     print(eth)
      4

c:\workspace\dpkt\dpkt\dpkt\ethernet.py in __init__(self, *args, **kwargs)
     77
     78     def __init__(self, *args, **kwargs):
---> 79         dpkt.Packet.__init__(self, *args, **kwargs)
     80         # if data was given in kwargs, try to unpack it
     81         if self.data:

[...]

PR #445, opened 2019-01-31
Handle zero Eth Type

kylekeppler:
This addresses issue #442. 

Proposed solution:

Check for `ethtype == 0` and handle as `Ethernet II` instead of `IEEE 802.3 Ethernet - LLC`. This is what Wireshark does.


kylekeppler:
Looks like I broke a LLC test. I'll look into if that was a valid test case originally.

obormot:
Good work. Thanks!

PR #449, opened 2019-04-20
Use `six` package for backwards compatibility

ngie-eign:
The `six` package provides a number of backwards compatible hooks for
moved/renamed APIs. It's better to use their APIs than reinvent the wheel.

While this doesn't completely remove the need for dpkt/compat.py, it
largely eliminates this need. Removing/replacing `compat_ord` and
remaining backwards compatible between python 2.7 and python 3.7 proved
to be more of a challenge than I was willing to commit to for this
change.

Signed-off-by: Enji Cooper <yaneurabeya@gmail.com>

### DIFF
--- a/dpkt/aoe.py
+++ b/dpkt/aoe.py
@@ -4,9 +4,10 @@ from __future__ import absolute_import
 
 import struct
 
+import six
+
 from . import dpkt
 from .decorators import deprecated
-from .compat import iteritems
 
 class AOE(dpkt.Packet):
     """ATA over Ethernet Protocol.
@@ -73,7 +74,7 @@ def __load_cmds():
     prefix = 'AOE_CMD_'
     g = globals()
 
-    for k, v in iteritems(g):
+    for k, v in six.iteritems(g):
         if k.startswith(prefix):
             name = 'aoe' + k[len(prefix):].lower()
             try:
--- a/dpkt/compat.py
+++ b/dpkt/compat.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 import sys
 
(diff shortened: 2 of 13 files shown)


ngie-eign:
Ping?

obormot:
I'm somewhat conflicted about this change. dpkt's philosophy is minimal to no dependencies. six is a 3rd party package, so it would introduce a dependency without adding functionality.

kbandla:
Using six came up when we were working on the migration to py3, and we decided to avoid it. If you notice, the compat module is a minimized six module, without importing the whole 3rd party module. 
Secondly, even if we take this approach, we do not entirely rid of the compat module anyway. So I agree with @obormot about introducing this module without really any added advantage. 
So unless we have a really strong case, Id close this PR. Hope you understand.

ngie-eign:
> I'm somewhat conflicted about this change, inclined toward rejecting it. dpkt's philosophy is minimal to no dependencies. six is a 3rd party package, so it would introduce a dependency without adding functionality.

six is a third party package, but it's one that's widely used in the open source community and although there might not be an immediate "win" from the change, it would be good in the longterm because it reduces the need for coming up with additional compatibility support and I'm guessing will be more widely tested than dpkt's compat.py, ensuring that if something subtle changes in the APIs, the six package is more likely to catch the issue sooner.

I will defer to the authors with this project, but my goal (when looking at this) was to reduce complexity when making FreeBSD tests function with py3, as the author who wrote the tests used dpkt.py as a standalone module for some opencrypto testing.

obormot:
cleanup.. consensus was to not merge this - closing

Issue #429, opened 2018-11-21
Add Travis CI Python 3.7 support


brifordwylie:
The Travis CI platform has some issues running Python 3.7 (this is not their fault but an OpenSSL on Trusty issue).. anyway figure out a way to have Travis run Python 3.7 tests.

Travis CI Python 3.7:
- https://github.com/travis-ci/travis-ci/issues/9815

Also see #428 for additional info

obormot:
Resolved by merging https://github.com/kbandla/dpkt/pull/450
Closing

PR #450, opened 2019-04-20
Remove python 2.6 support and add python 3.7 test coverage

ngie-eign:
## About 2.6:

python 2.6 has been unsupported for at least 4 years. Drop support for
the python version as there have been several changes made in recent
years to fix functional and security bugs with the interpreter, etc.

It's best to upgrade to python 2.7 (if one is stuck on 2.x), or upgrade
to py3.

## About 3.7:

python 3.7.0 was released 10 months ago. As such, it's reasonable to add
test coverage to ensure it functions.

It is reasonably close feature-wise to 3.6 that there will unlikely be
any behavioral difference in this package. Some of the benefits of 3.7
over earlier versions is improved `asyncio` module support and general
performance improvements.

More information about what's new in 3.7 can be found
[here](https://docs.python.org/3.7/whatsnew/3.7.html).

## About the Travis changes:

* Switch from Trusty (the default Ubuntu distro) to Xenial in order to support
  python 3.7 testing with Travis, as Trusty lacks python 3.7 support.
* Use pypy3.5 since the pypy package (unversioned) isn't available for Xenial,
  per the [Travis CI docs](https://docs.travis-ci.com/user/languages/python/#pypy-support).

## About `dpkt.decorators`:

As part of this work, I realized that `dpkt.decorators` was unused. As such,
reap all references to the module and the module itself.


ngie-eign:
Ping?

obormot:
this would also close #429  
looks like Travis is happy w/ py3.7 now. cool

PR #455, opened 2019-07-12
fixed ipv6 packet so that it can be used for generating IPv6 data

joakimeriksson:
The IPv6 packet in dpkt was broken and had some code that was not working in the IPv6 case - assuming an updated p (proto) variable. This will fix that plus also ensure that if there are no extension headers it will still generate a packet - and set p to whatever is the last next-header value.

### DIFF
--- a/dpkt/ip6.py
+++ b/dpkt/ip6.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 
 from . import dpkt
 from . import ip
+from . import tcp
 from .compat import compat_ord
 
 
@@ -27,7 +28,6 @@ class IP6(dpkt.Packet):
         ('src', '16s', ''),
         ('dst', '16s', '')
     )
-
     _protosw = ip.IP._protosw
 
     @property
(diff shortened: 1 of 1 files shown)


joakimeriksson:
Hello - any action here?

dpkt/ip6.py
```diff
@@ -91,29 +91,34 @@ def unpack(self, buf):
...
+            nxt = self.all_extension_headers[-1].nxt
+            return nxt, b''.join(bytes(ext) for ext in self.all_extension_headers)
         # Output extension headers in order defined in RFC1883 (except dest opts)
         header_str = b""
-        for hdr in ext_hdrs:
-            if hdr in self.extension_hdrs:
-                header_str += bytes(self.extension_hdrs[hdr])
-        return header_str
+        if hasattr(self, 'extension_hdrs'):
+            for hdr in ext_hdrs:
```
obormot: While at it, could you possibly move ```ext_hdrs``` and ```ext_hdrs_cls``` definitions higher in the file, and make them uppercase (as they are global constants)?

dpkt/ip6.py
```diff
@@ -91,29 +91,34 @@ def unpack(self, buf):
...
+                    header_str += bytes(self.extension_hdrs[hdr])
+        return nxt, header_str
 
     def __bytes__(self):
+        self.p, hdr_str = self.headers_str()
         if (self.p == 6 or self.p == 17 or self.p == 58) and not self.data.sum:
             # XXX - set TCP, UDP, and ICMPv6 checksums
             p = bytes(self.data)
-            s = dpkt.struct.pack('>16s16sxBH', self.src, self.dst, self.nxt, len(p))
+            s = dpkt.struct.pack('>16s16sxBH', self.src, self.dst, self.p, len(p))
```
obormot: hmm, this should probably be just ```struct.pack```

dpkt/ip6.py
joakimeriksson: Ok, I will move them up!

joakimeriksson: I moved one of them up now - the CLS one needs the classes defined before being assigned. All is uppercase now at least.

dpkt/ip6.py
joakimeriksson: yes, that is probably a valid comment also for the IP one. I fixed this now.

Issue #385, opened 2017-08-31
Python3 OSPF default auth field packet packing error


nfz1:
There is a bug in the current version of dpkt on Python3. If the auth header field is omitted when initiating an OSPF packet, an exception is raised. Here is a quick test and stack-trace using the Python3 interpreter.

```Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from dpkt.ospf import OSPF
>>> packet = OSPF(v=1,type=1,area=0,len=0,router=0,data=b'')
>>> packet.__bytes__()
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/dpkt/dpkt.py", line 150, in pack_hdr
    *[getattr(self, k) for k in self.__hdr_fields__])
struct.error: argument for 's' must be a bytes object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/dpkt/dpkt.py", line 160, in pack_hdr
    return struct.pack(self.__hdr_fmt__, *vals)
struct.error: argument for 's' must be a bytes object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/dpkt/ospf.py", line 36, in __bytes__
    self.sum = dpkt.in_cksum(dpkt.Packet.__bytes__(self))
  File "/usr/local/lib/python3.5/dist-packages/dpkt/dpkt.py", line 144, in __bytes__
    return self.pack_hdr() + bytes(self.data)
  File "/usr/local/lib/python3.5/dist-packages/dpkt/dpkt.py", line 162, in pack_hdr
    raise PackError(str(e))
dpkt.dpkt.PackError: argument for 's' must be a bytes object
```

The problem is due to the auth field having an invalid value set as its default `''`. This should be `b''` in python 3 as strings are now stored in Unicode. Can also fix this by using the encode method to translate the string to bytes before the packet processing occurs, but this may be more messy.

Here is a link to the file and line in question. [dpkt/ospf.py line 31](https://github.com/kbandla/dpkt/blob/master/dpkt/ospf.py#L31)

If we set the auth field value to `b''` we get the expected output.

```Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[...]

PR #461, opened 2019-10-23
Turn strings in headers into bytes. Necessary for Py3 compatibility.

mx:
I was performing a Python3 migration of code that depends on dpkt when I found a problem with bytes(packet) or str(packet). The header packing methods would throw with the following error:
"struct.error: argument for 's' must be a bytes object"

Some other github issues have been created on this same topic for specific packet types; I went through and fixed every example that I could find. This should be a NOP in Python2 (as b'' is an alias for '' there), and fix the above errors in Python3.


dpkt/hsrp.py
```diff
@@ -39,5 +39,5 @@ class HSRP(dpkt.Packet):
         ('group', 'B', 0),
         ('rsvd', 'B', 0),
         ('auth', '8s', 'cisco'),
```
obormot: ```b'cicso'``` ?

obormot:
Good stuff. I'm surprised the unit tests didn't catch any of this. Please add at least one unit test that would fail before the change and work after the change. 

> Some other github issues have been created on this same topic

Please reference these here so we could close them by merging this PR

mx:
> Please reference these here so we could close them by merging this PR
#385 

> Please add at least one unit test that would fail before the change and work after the change.
Done

mx:
Not sure how it thinks coverage *decreased* when I added that test, but I'm gonna guess that's a false positive.

dpkt/ethernet.py
```diff
@@ -717,6 +717,16 @@ def test_eth_802dot1ad_802dot1q_ip():
     assert str(e1) == str(e2)
 
 
+def test_eth_pack():
```
obormot: indentation needs to be 4 spaces

mx: Done.

obormot:
> Not sure how it thinks coverage _decreased_ when I added that test, but I'm gonna guess that's a false positive.

It decreased because the 2 lines you added in the unit test
```
except:
  raised = True
```
were not executed hence decreased coverage (dumb I know..)

Perhaps restructure the unit test into something like
```
eth = Ethernet(data=b'12345')
assert str(eth)
```

mx:
> > Not sure how it thinks coverage _decreased_ when I added that test, but I'm gonna guess that's a false positive.
> 
> It decreased because the 2 lines you added in the unit test
> 
> ```
> except:
>   raised = True
> ```
> 
> were not executed hence decreased coverage (dumb I know..)
> 
> Perhaps restructure the unit test into something like
> 
> ```
> eth = Ethernet(data=b'12345')
> assert str(eth)
> ```

This feels less clear to the viewer than what's going on there now. I can do something like that if you want, but it feels like making it worse in order to please a metric (coverage).

obormot:
> This feels less clear to the viewer than what's going on there now. I can do something like that if you want, but it feels like making it worse in order to please a metric (coverage).

It's rather to satisfy the metric in order to be able to merge. Can't merge it with decreased coverage

mx:
> > This feels less clear to the viewer than what's going on there now. I can do something like that if you want, but it feels like making it worse in order to please a metric (coverage).
> 
> It's rather to satisfy the metric in order to be able to merge. Can't merge it with decreased coverage

Done, then.

mx:
The remaining coverage failure is because I added the test to main with the other tests.

Issue #465, opened 2019-12-30
IEEE 802.1Q double tagging (QinQ) with miscellaneous data parsing problem


EliScada:
In ethernet module in case of vlan double tagging with a following unknown data part (isinstance(self.data, bytes)==True) the first 4 bytes of the data are parsed as an additional vlan tag.

This happens because when self.data is not of a particular known type, _unpack_data is being called once more. Since the next_type variable is local and is given the value of the uppermost tag (which would be vlan) the code enters an additional iteration of the loop that appends vlan_tags and bites into the first 4 bytes of the data.

EliScada:
The following unit test illustrates the issue:
```
def test_eth_802dot1q_with_unfamiliar_data():
    profinet_data = (b'\xfe\xff\x05\x01\x05\x01\x00\x02\x00\x00\x00\x6c\x02'
         b'\x05\x00\x12\x00\x00\x02\x01\x02\x02\x02\x03\x02\x04\x02\x05\x02'
         b'\x06\x01\x01\x01\x02\x02\x01\x00\x08\x00\x00\x53\x37\x2d\x33\x30'
         b'\x30\x02\x02\x00\x22\x00\x00\x70\x6c\x63\x78\x62\x33\x30\x30\x78'
         b'\x6b\x63\x70\x75\x78\x61\x33\x31\x37\x2d\x32\x78\x61\x70\x6e\x78'
         b'\x72\x64\x70\x32\x32\x63\x66\x02\x03\x00\x06\x00\x00\x00\x2a\x01'
         b'\x01\x02\x04\x00\x04\x00\x00\x02\x00\x01\x02\x00\x0e\x00\x01\xc0'
         b'\xa8\x3c\x87\xff\xff\xff\x00\xc0\xa8\x3c\x87')

    s = (b'\x00\x0c\x29\x65\x1c\x29\x00\x0e\x8c\x8a\xa2\x5e\x81\x00\x00\x00'
         b'\x88\x92' + profinet_data)

    eth = Ethernet(s)
    assert eth.type == ETH_TYPE_8021Q
    assert len(eth.vlan_tags) == 1
    assert eth.vlan_tags[0].type == ETH_TYPE_PROFINET
    assert isinstance(eth.data, bytes)
    assert eth.data == profinet_data
```

EliScada:
I fixed the issue by saving the next_type variable on a class instance level.

EliScada:
[...]

PR #466, opened 2019-12-30
Fix QinQ vlan tag parsing with miscellaneous data

EliScada:
Fix issue https://github.com/kbandla/dpkt/issues/465

Change scope of next_type variable.

Add unit test: 
 1. Verify nested eth_type
 2. Data part left intact

### DIFF
--- a/dpkt/ethernet.py
+++ b/dpkt/ethernet.py
@@ -126,7 +126,7 @@ class Ethernet(dpkt.Packet):
                 self._next_type = ETH_TYPE_TEB  # re-use TEB class mapping to decode Ethernet
 
         try:
-            eth_type = self._next_type if self._next_type else self.type
+            eth_type = self._next_type or self.type
             self.data = self._typesw[eth_type](buf)
             setattr(self, self.data.__class__.__name__.lower(), self.data)
         except (KeyError, dpkt.UnpackError):
--- a/dpkt/ethernet.py
+++ b/dpkt/ethernet.py
@@ -52,6 +52,7 @@ ETH_TYPE_PPPoE_DISC = 0x8863  # PPP Over Ethernet Discovery Stage
 ETH_TYPE_PPPoE = 0x8864  # PPP Over Ethernet Session Stage
 ETH_TYPE_LLDP = 0x88CC  # Link Layer Discovery Protocol
 ETH_TYPE_TEB = 0x6558  # Transparent Ethernet Bridging
+ETH_TYPE_PROFINET = 0x8892  # PROFINET protocol
 
 # all QinQ types for fast checking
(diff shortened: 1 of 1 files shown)


EliScada:
Hi @obormot I saw you have a thorough acquaintance with the ethernet module, would be great if you could find the time to have a look.
Thanks Eli

dpkt/ethernet.py
```diff
@@ -116,16 +117,17 @@ def _unpack_data(self, buf):
...
             # pseudowire Ethernet
             elif len(buf) >= self.__hdr_len__:
                 if buf[:2] == b'\x00\x00':  # looks like the control word (ECW)
                     buf = buf[4:]  # skip the ECW
-                next_type = ETH_TYPE_TEB  # re-use TEB class mapping to decode Ethernet
+                self._next_type = ETH_TYPE_TEB  # re-use TEB class mapping to decode Ethernet
 
         try:
-            self.data = self._typesw[next_type](buf)
+            eth_type = self._next_type if self._next_type else self.type
```
obormot: (minor) could use `eth_type = self._next_type or self.type` as a shorter form

dpkt/ethernet.py
```diff
@@ -77,31 +78,31 @@ class Ethernet(dpkt.Packet):
...
     def __init__(self, *args, **kwargs):
+        self._next_type = None
         dpkt.Packet.__init__(self, *args, **kwargs)
         # if data was given in kwargs, try to unpack it
         if self.data:
             if isstr(self.data) or isinstance(self.data, bytes):
                 self._unpack_data(self.data)
 
     def _unpack_data(self, buf):
-        next_type = self.type
```
obormot: is it cool to remove this completely?  instead of `self._next_type = self.type` 
I see this on line 139 now under `if self.type > 1500:`.  Just wanted to make sure it's the intended and explainable change

EliScada: Hey, moving next type assignment to line 139 is intended - in addition to moving it I've also changed it's scope from local to `_unpack_data()` function to class instance variable. 
`_unpack_data()` is called potentially 2 times - first from within `unpack()` but also from `__init__()` line 86 in case `self.data` is of an unfamiliar type. In the latter case, we want the "next type" to be saved from the previous run of `_unpack_data()` as it points to the last eth type encountered as opposed to the self.type which saves the most upper-level eth type, which would be one of `_ETH_TYPES_QINQ`.

dpkt/ethernet.py
EliScada: Yep, nice

EliScada:
Thanks for the review @obormot

Issue #458, opened 2019-09-11
Netbios name encoding bug


EvanSonnemans:
When encoding names in netbios packets I encounter the following problem:
```
Python 3.7.4 (default, Jul  9 2019, 18:13:23)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import dpkt
>>> payload = bytes.fromhex(
...     '20434b4141414141414141414141414141'
...     '4141414141414141414141414141414100')
>>> decoded_payload = dpkt.netbios.decode_name(payload)
>>> print(decoded_payload)
b' CKAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\x00'
>>> encoded_payload = dpkt.netbios.encode_name(decoded_payload)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/dpkt/netbios.py", line 16, in encode_name
    c = ord(c)
TypeError: ord() expected string of length 1, but int found
>>>
```
 
It seems like dpkt has a bug in encode_name and decode_name where python2 is not correctly converted to python3. This is my proposed fix:

```
import struct
import dpkt
import binascii


def encode_name(decoded_name):
    """Return the NetBIOS first-level encoded name."""
    encoded_name = bytearray()
    for c in decoded_name:
        encoded_name.append((c >> 4) + 0x41)
        encoded_name.append((c & 0xf) + 0x41)
    return encoded_name


def decode_name(encoded_name):
    """Return the NetBIOS first-level decoded nbname."""
    decoded_name = bytearray()
    for i in range(0, len(encoded_name), 2):
        decoded_name.append(
            ((encoded_name[i] - 0x41) << 4) |
            ((encoded_name[i + 1] - 0x41) & 0xf))
    return decoded_name[:decoded_name.index(b'\x00')+1]


payload = bytes.fromhex(
    '20434b4141414141414141414141414141'
    '4141414141414141414141414141414100')
print(payload)
decoded_payload = encode_name(payload)
print(decoded_payload)
encoded_payload = decode_name(decoded_payload)
print(payload, encoded_payload)
print(payload == encoded_payload)
print('Great success!!!')
```
I have no idea why there is a character limit of 16 chars, I have removed them in my fix, if this is protocol related they have to be reinserted.

obormot:
[...]

PR #469, opened 2020-03-31
fix netbios name encoding and decoding

obormot:
fix for #458 
netbios module needs unit tests


obormot:
All Travis builds pass
https://travis-ci.org/github/kbandla/dpkt/builds/669103427

Issue #438, opened 2018-12-20
dpkt.Writer does not write packet trailer


mikewalshchicago:
I'm new to GitHub so I apologize if this is not the correct place to ask questions, but i could not find anything regarding packet trailers in the documentation.

I am using pcapy.open_live() to read the packets from a bytestream.  I am then using dpkt to do some read-only analysis, and then writing each packet to a pcap file using dpkt.Writer.  An abstract of what my code is doing is as follows:

#first, i'm opening the bytestream, passing the desired interface from cli arguments.  I have tried larger #snaplen values however I saw no difference 
cap = pcapy.open_live(interface, 65536, 1, 0)

#i want to write every UDP message in this packet capture, so i do the following:
while True:
(header, packet) = cap.next()
ethernet = dpkt.ethernet.Ethernet(packet)

   if ethernet.type == dpkt.ethernet.ETH_TYPE_IP:
      ip = ethernet.data
      if ip.p == dpkt.ip.IP_PROTO_UDP:
          udp = ip.data
#there's some processing that goes on in the middle, but i am not manipulating the output. I have tried #different snaplen arguments here as well, e.g. snaplen=65536, but see no difference in the output
   writer = dpkt.pcap.Writer(open(path + filename)+'.pcap', 'wb+'),nano=True)
#I have also tried to write 'ip' as ouput, which appears the same.  
   writer.writepkt(ethernet)

The headers and payloads look fine in my pcap, but each message also has an 16 byte packet trailer which i do not see. if I use tcpdump -w, no special arguments are needed to see the trailer. 

I'd appreciate any help you can give me, and happy to help where i can with diagnosis

obormot:
Please attach a sample pcap - even 1 packet with the 16-byte trailer would help to reproduce the issue. I think dpkt may not be preserving the padding / trailer in some cases.

mikewalshchicago:
Please see attached, and thanks for the quick reply!

[trailer.zip](https://github.com/kbandla/dpkt/files/2705353/trailer.zip)

obormot:
Confirmed. Interestingly enough there's a single IP packet in the pcap and exactly this packet gets packaged w/o the 16 bytes trailer!
```
In [31]: str(eth) == pp[1]
Out[31]: False

In [32]: len( str(eth) )
Out[32]: 243

In [33]: len(pp[1])
Out[33]: 259

# difference is 16 bytes
```

[...]

PR #470, opened 2020-04-05
fix 802.3 Ethernet fcs, padding and trailer

obormot:
#438 had an example pcap which revealed that dpkt.ethernet was incorrectly decoding some FCS/padding/trailer bytes. This PR addresses these issues. Since some of it is heuristic I've looked at comments in Wireshark src `https://github.com/wireshark/wireshark/blob/master/epan/dissectors/packet-eth.c : add_ethernet_trailer()`


obormot:
@brifordwylie while you're at it, could you please review/merge this PR?

brifordwylie (approved):

Issue #362, opened 2017-04-26
dpkt pcap next fails in python 2.7


bmloveridge:
I'm using dpkt in python 2.7. Initially wrote my code with dpkt 1.8.8 and used next() on the pcap Reader class to get the next packet. Recently updated to 1.9.1 and this code broke complaining there is no next() method for the Reader class. Found that __next__ method is defined and switching code to call that directly which allowed code to run. So its seems dpkt 1.9.1 will work with Python 3 but not Python 2.7. For now I've switched back to 1.8.8

obormot:
@bmloveridge code example please

kbandla:
@bmloveridge any updates? I am working on a patch for the Reader class, and wanted to get this issue out of the way as well.

bmloveridge:
Don't have access to a computer for the next week to create a simple
example of the fail that I can demonstrate shows the exact fail.

Code would be really simple though. Simply open the stream with dpkt and
then use .next() to get data from stream. This fails in python 2.7. But if
I change the .next() to .__next__() it works.

Switching back to 1.8.8 allows .next() to get next data packet.

obormot:
cleaning up old issues..  looks like `Reader.next()` was replaced with`.__next__()` as part of python3 migration in 1.9.1. 
https://github.com/kbandla/dpkt/commit/433cd5dc520cf0f03a63f6eb005d65421c894521#diff-df1c085b0dae8fc5c31b27cffd336d24L273
Trivial to migrate to calling `next(reader_instance)` => closing this

PR #473, opened 2020-04-23
Fix compatibility Python 2-3

shadchin:


obormot:
linking related issue #362

PR #479, opened 2020-07-11
ipv6 option unpack: use byte length in the loop

crosser:
self.length is the _bit_ length of the OptsHeader, while in the while loop, we need _byte_ length.

Attempt to continue the loop beyond the data size eventually results in IndexError exception.

This should fix the issue #477


obormot:
I believe ip6 parser does the right thing. This proposed change breaks existing unit tests.

PR #480, opened 2020-07-13
BGP: Handle multiple capabilities in one Optional Parameter

### LINKED ISSUE #118: BGP (v4) Open typed packet failing to parse.
_From [Mawr...@gmail.com](https://code.google.com/u/110708298735800734951/) on June 22, 2012 11:38:15_

What steps will reproduce the problem? 1. Packet received was:

\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\x00'\x01\x04\xfd\xe8\x00\xb4\xc0\xa8\x02\x01\n\x02\x08\x01\x04\x00\x01\x00\x01\x02\x00

call dpkt.bgp.BGP(buf)

Raises exception:

  File "bgp_dump.py", line 11, in \<module>
    p = bgp.BGP(data)
  File "/usr/local/lib/python2.7/site-packages/dpkt/dpkt.py", line 75, in __init__
    self.unpack(args[0])
  File "/usr/local/lib/python2.7/site-packages/dpkt/bgp.py", line 130, in unpack
    self.data = self.open = self.Open(self.data)
  File "/usr/local/lib/python2.7/site-packages/dpkt/dpkt.py", line 75, in __init__
    self.unpack(args[0])
  File "/usr/local/lib/python2.7/site-packages/dpkt/bgp.py", line 157, in unpack
    param = self.Parameter(self.data)
  File "/usr/local/lib/python2.7/site-packages/dpkt/dpkt.py", line 75, in __init__
    self.unpack(args[0])
  File "/usr/local/lib/python2.7/site-packages/dpkt/bgp.py", line 188, in unpack
    self.data = self.capability = self.Capability(self.data)
  File "/usr/local/lib/python2.7/site-packages/dpkt/dpkt.py", line 78, in __init__
    raise NeedData
dpkt.dpkt.NeedData What is the expected output? What do you see instead? The expected output is that the packet is parsed, with the ROUTE REFRESH capability enabled. What version of the product are you using? On what operating system? # $Id: bgp.py 52 2008-08-25 22:22:34Z jon.oberheide $

On Ubuntu, but I don't think it's a platform dependent issue. Please provide any additional information below. I believe the cause to be is if the packet contains a zero-length capabilities (route refresh), the resulting data is 0, but it tries to unpack it anyways, causing the exception.

According to the RFC ( http://tools.ietf.org/html/rfc2918 ) the zero-length capabilities is expected.  

A solution I came up with was to check if self.data is length 0, and if so, return before calling unpack.  Added this to line 183 in bgp.py:

                if len(self.data) == 0:
                    return

_Original issue: http://code.google.com/p/dpkt/issues/detail?id=91_

kbandla:
* Fixes #118 (please see issue for details)
* Introduce `capabilities` property (for cases where there are >1 capablity)
* Keep older `capability` property for backward compatibility
* Added `__bgp10` unittest, which broke the original parser

### DIFF
--- a/dpkt/bgp.py
+++ b/dpkt/bgp.py
@@ -77,6 +77,9 @@ CAPABILITY = 2
 # Capability Types
 CAP_MULTIPROTOCOL = 1
 CAP_ROUTE_REFRESH = 2
+CAP_GRACEFUL_RESTART = 64
+CAP_SUPPORT_4OCTETAS = 65
+CAP_ADD_PATH = 69
 
 # NOTIFICATION Error Codes
 MESSAGE_HEADER_ERROR = 1
@@ -200,7 +203,22 @@ class BGP(dpkt.Packet):
                 if self.type == AUTHENTICATION:
                     self.data = self.authentication = self.Authentication(self.data)
                 elif self.type == CAPABILITY:
-                    self.data = self.capability = self.Capability(self.data)
+                    l = []
+                    while self.data:
+                        capability = self.Capability(self.data)
+                        l.append(capability)
+                        self.data = self.data[self.__hdr_len__+capability.len:]
+                    self.data = self.capabilities = l
+                    # define capability to keep backward compatibility
+                    self.capability = self.capabilities[0]
+
+            def __len__(self):
+                return self.__hdr_len__ + sum(map(len, self.capabilities))
+
+            def __bytes__(self):
+                caps = b''.join(map(bytes, self.capabilities))
+                self.cap_len = len(caps)
+                return self.pack_hdr() + caps
 
             class Authentication(dpkt.Packet):
                 __hdr__ = (
@@ -216,6 +234,7 @@ class BGP(dpkt.Packet):
                 def unpack(self, buf):
                     dpkt.Packet.unpack(self, buf)
                     self.data = self.data[:self.len]
+
 
     class Update(dpkt.Packet):
         __hdr_defaults__ = {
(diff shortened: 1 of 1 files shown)


dpkt/bgp.py
```diff
@@ -200,7 +203,22 @@ def unpack(self, buf):
...
+                    self.data = self.capabilities = l
+                    # define capability to keep backward compatibility
+                    self.capability = self.capabilities[0]
+
+            def __len__(self):
+                return self.__hdr_len__ + sum(map(len, self.capabilities))
+
+            def __bytes__(self):
+                caps = b''.join(map(bytes, self.capabilities))
+                self.cap_len = len(caps)
```
obormot, line 220: unused?
also prob shouldn't create new fields inside this method

dpkt/bgp.py
```diff
@@ -200,7 +203,22 @@ def unpack(self, buf):
...
+                    while self.data:
+                        capability = self.Capability(self.data)
+                        l.append(capability)
+                        self.data = self.data[self.__hdr_len__+capability.len:]
+                    self.data = self.capabilities = l
+                    # define capability to keep backward compatibility
+                    self.capability = self.capabilities[0]
+
+            def __len__(self):
+                return self.__hdr_len__ + sum(map(len, self.capabilities))
```
obormot, line 216: calling `len()` on each instance of capability would cause serialization to bytes, then taking the resulting length.  would something like `sum(cap.len for cap in self.capabilities)` achieve the same result?

dpkt/bgp.py
```diff
@@ -200,7 +203,22 @@ def unpack(self, buf):
                 if self.type == AUTHENTICATION:
                     self.data = self.authentication = self.Authentication(self.data)
                 elif self.type == CAPABILITY:
-                    self.data = self.capability = self.Capability(self.data)
+                    l = []
+                    while self.data:
+                        capability = self.Capability(self.data)
+                        l.append(capability)
+                        self.data = self.data[self.__hdr_len__+capability.len:]
```
obormot, line 210: I believe pep8 would want whitespace around `+`

amgadhanafy:
Hi @kbandla ,
Any chance to get this merged soon?

dpkt/bgp.py
crocogorical, line 216: This doesn't cover the case where the type is `AUTHENTICATION`, as `capabilities` won't exist (and it doesn't count the length of the authentication data).

dpkt/bgp.py
```diff
@@ -200,7 +203,22 @@ def unpack(self, buf):
...
+                    self.data = self.capabilities = l
+                    # define capability to keep backward compatibility
+                    self.capability = self.capabilities[0]
+
+            def __len__(self):
+                return self.__hdr_len__ + sum(map(len, self.capabilities))
+
+            def __bytes__(self):
+                caps = b''.join(map(bytes, self.capabilities))
```
crocogorical, line 219: This will also fail if this is an authentication parameter. Perhaps some variant of `caps = bytes(self.data) if not isinstance(self.data, list) else b''.join(map(bytes, self.data))`?

dpkt/bgp.py
crocogorical: Alternatively you could override the `__len__` method to return that header element?

PR #487, opened 2020-07-25
BGP: add support for MPReachNLRI next hop list and IP6

### LINKED ISSUE #485: BGP: invalid length of packed IP address string (error reading next hop when it carry more than one IP)
Data set from https://packetlife.net/captures/protocol/bgp/ 
https://packetlife.net/media/captures/BGP_MP_NLRI.cap 
Packet 20 

next-hop has 2 ip6 (32 bytes) but considered as one IP

error
```
ERROR - __main__ - extract_features:88 - Error processing BGP data: packet 20 : message 1 of BGP_MP_NLRI.cap
ERROR - __main__ - extract_features:89 - BGP_data(108): b'\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\x00\x6c\x02\x00\x00\x00\x55\x40\x01\x01\x00\x40\x02\x04\x02\x01\xfd\xe9\x80\x04\x04\x00\x00\x00\x00\x80\x0e\x40\x00\x02\x01\x20\x20\x01\x0d\xb8\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\xfe\x80\x00\x00\x00\x00\x00\x00\xc0\x01\x0b\xff\xfe\x7e\x00\x00\x00\x40\x20\x01\x0d\xb8\x00\x01\x00\x02\x40\x20\x01\x0d\xb8\x00\x01\x00\x01\x40\x20\x01\x0d\xb8\x00\x01\x00\x00'
ERROR - __main__ - extract_features:90 - invalid length of packed IP address string
  File "C:\Users\Amgad\PycharmProjects\bell\bgp_update.py", line 265, in inet_to_str
    return socket.inet_ntop(socket.AF_INET, inet)
ValueError: invalid length of packed IP address string

During handling of the above exception, another exception occurred:

  File "C:\Users\Amgad\PycharmProjects\bell\bgp_update.py", line 81, in extract_features
    BgpFeatureExtractor.extract_update_attributes(update)
  File "C:\Users\Amgad\PycharmProjects\bell\bgp_update.py", line 207, in extract_update_attributes
    logger.debug("\t\t\t\tnext_hop: " + inet_to_str(bytes(hop)))
  File "C:\Users\Amgad\PycharmProjects\bell\bgp_update.py", line 267, in inet_to_str
    return socket.inet_ntop(socket.AF_INET6, inet)
ValueError: invalid length of packed IP address string
```
packet in wire shark 
image

amgadhanafy:
- Fixes #485 (please see issue for details)
- Introduce next_hops property (list of next_hop )
- Add support for IP6 MPReachNLRI next hop
- Keep older next_hop property for backward compatibility 
- Added test_bgp_mp_nlri_20_1_mp_reach_nlri_next_hop unittest


dpkt/bgp.py
```diff
@@ -935,7 +943,96 @@ def test_unpack():
...
+    assert (inet_to_str(mp_reach_nlri.next_hop[1]) == 'fe80::c001:bff:fe7e:0')
+
+
 if __name__ == '__main__':
     test_pack()
     test_unpack()
+    test_bgp_mp_nlri_20_1_mp_reach_nlri_next_hop()
     print('Tests Successful...')
+
+def inet_to_str(inet):
```
obormot, line 1026: A bit confusing to see this def here. I'd rather put it inside the `test_bgp_mp_nlri_20_1_mp_reach_nlri_next_hop()` body since it's the only place it's used.

obormot:
> * Introduce next_hop property (for cases where there are >1 next_hop )
> ...
> * Keep older next_hop property for backward compatibility

Looking at the diff I think this description needs to be updated (e.g. the existing `next_hop` attribute is changed into a list, breaking backward compatibility).

dpkt/bgp.py
```diff
@@ -511,10 +511,18 @@ def unpack(self, buf):
...
                     # Next Hop
+                    hop_len = 4
+                    if self.afi == AFI_IPV6:
+                        hop_len = 16
+                    l = []
                     nlen = struct.unpack('B', self.data[:1])[0]
                     self.data = self.data[1:]
-                    self.next_hop = self.data[:nlen]
-                    self.data = self.data[nlen:]
+                    num_hops = nlen // hop_len
```
obormot, line 520: I'm worried that using `//` here (instead of `/`) may mask some weird issue when the packet is misconstructed etc.  If it's supposed to divide as integer I'd prefer to see regular division here.

dpkt/bgp.py
amgadhanafy: inlined

amgadhanafy:
> > * Introduce next_hop property (for cases where there are >1 next_hop )
> >   ...
> > * Keep older next_hop property for backward compatibility
> 
> Looking at the diff I think this description needs to be updated (e.g. the existing `next_hop` attribute is changed into a list, breaking backward compatibility).

fixed

dpkt/bgp.py
amgadhanafy: replaced with while loop instead of division

dpkt/bgp.py
```diff
@@ -511,10 +511,19 @@ def unpack(self, buf):
                     dpkt.Packet.unpack(self, buf)
 
                     # Next Hop
+                    hop_len = 4
+                    if self.afi == AFI_IPV6:
+                        hop_len = 16
+                    l = []
                     nlen = struct.unpack('B', self.data[:1])[0]
                     self.data = self.data[1:]
                     self.next_hop = self.data[:nlen]
```
obormot, line 520: please add a comment explaining that `next_hop` is kept for backward compatibility

dpkt/bgp.py
```diff
@@ -935,7 +944,83 @@ def test_unpack():
     assert (r.mpls_label_stack == b'\x00\x00\x02\x00\x00\x00')
 
 
+def test_bgp_mp_nlri_20_1_mp_reach_nlri_next_hop():
+    # Error processing BGP data: packet 20 : message 1 of BGP_MP_NLRI.cap
```
obormot, line 948: Since the error will be fixed by this PR, it's better to just refer the github issue in the comment here (https://github.com/kbandla/dpkt/issues/485)

dpkt/bgp.py
amgadhanafy: added

dpkt/bgp.py
amgadhanafy: added github issue reference

Issue #498, opened 2020-12-10
Parsing truncated DNS packets


brightpinefield:
I'm trying to dump DNS data out of pcaps that are truncated to the first 100 bytes.  Found a script here https://mmishou.wordpress.com/2010/04/13/passive-dns-mining-from-pcap-with-dpkt-python/ to test dpkt but it seems to not go past the  "try: dns = dpkt.dns.DNS(udp.data)" .. it hits the continue and I get nothing.. is it possible to use dpkt with truncated packets?  I have data in the DNS packet (see screenshot) that I was hoping to get at..

![Screen Shot 2020-12-10 at 2 06 19 PM](https://user-images.githubusercontent.com/62759487/101817792-e98fa780-3af0-11eb-96c2-95ff8415405b.png)

kbandla:
@brightpinefield would it be possible for you to share the pcap? I suspect that lib will fail with `dpkt.UnpackError`, but wanted to check.

brightpinefield:
@kbandla  Sure. I've attached it..Thx for the help!
[test.pcap.zip](https://github.com/kbandla/dpkt/files/5675256/test.pcap.zip)

kbandla:
Yes, there is way to parse truncated data! 
We do this by first creating a `dns` object and then calling the `unpack` method, with the UDP data. This will populate the fields in the dns object as far as data in the packet is available, till an exception is raised.  
However note that `print(dns_object)` function will fail because it will internally call the unpack routine again, leading to failure. 

Example:
```python
....
for ts, buf in pcap:
    try:
        eth = dpkt.ethernet.Ethernet(buf)
        if eth.type != 2048:
            continue
        ip = eth.data
        if ip.p != 17:
            continue
        udp = ip.data
        if udp.sport != 53 and udp.dport != 53:
            continue
        dns = dpkt.dns.DNS()        # <---- THIS
        dns.unpack(udp.data)        # <---- HERE
    except Exception as e:
        print('ErrorParsingDNS: %s'%e)
        print('Warning: Not all DNS fields might have been populated')
    
    if dns.qr != dpkt.dns.DNS_R:
        continue
    if dns.opcode != dpkt.dns.DNS_QUERY:
        continue
    if dns.rcode != dpkt.dns.DNS_RCODE_NOERR:
        continue
    if len(dns.an) < 1:
        continue
    
    for answer in dns.an:
        if answer.type == 5:
            print('CNAME: %s - %s'%(answer.name, answer.cname))
[...]

PR #508, opened 2020-12-23
Truncated packet example

brifordwylie:


brifordwylie:
Example script for #498

PR #509, opened 2020-12-27
fix SCTP decoder DATA chunk padding and handling of .data field

### LINKED ISSUE #499: Issue with SCTP DATA chunk padding
Hello,
while using dpkt, I identified an issue in SCTP DATA chunk management.
An SCTP DATA chunk must be padded with 0, in order to align 4-bytes boundary. Fix may be located in class Chunk:

```python
class Chunk(dpkt.Packet):
    __hdr__ = (
        ('type', 'B', INIT),
        ('flags', 'B', 0),
        ('len', 'H', 0)
    )

    def unpack(self, buf):
        dpkt.Packet.unpack(self, buf)
        # FIX: SCTP DATA Chunked is padded, 4-bytes aligned
        length = self.len
        if length % 4 > 0:
            length = length + 4 - (length % 4)
        self.data = self.data[:length - self.__hdr_len__]
```

I assume your fix would be smarter than my proposal

obormot:
This fixes #499 and adds a unit test based on a pcap I found among Wireshark sample captures.

While working on the fix I noticed SCTP decoder incorrectly uses its `.data` field, basically copying `.chunks` as seen in this line:
```
self.data = self.chunks = l
```
All other dpkt decoders utilize the `.data` for any bytes remaining after decoding all the known fields (header fields and `.chunks` in this case), leaving it for other decoders to pick up if necessary. I changed the sctp decoder to align with this convention and use `.chunks` instead of `.data` when packing and unpacking. All existing unit tests pass without changes.


brifordwylie (approved):
👍

Issue #505, opened 2020-12-14
Seven tests fail on s390x


mcepl:
When building package for dpkt for OpenSUSE/Factory, I get these tests to fail on [s390x](https://en.wikipedia.org/wiki/Linux_on_IBM_Z#Hardware), which has very weird hardware. 
[Complete build log](https://github.com/kbandla/dpkt/files/5691397/_log.txt)

```
[   27s] =================================== FAILURES ===================================
[   27s] __________________________________ test_utils __________________________________
[   27s] 
[   27s]     def test_utils():
[   27s]         __buf = b'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e'
[   27s]         __hd = '  0000:  00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e     ...............'
[   27s]         h = hexdump(__buf)
[   27s]         assert (h == __hd)
[   27s]         c = in_cksum(__buf)
[   27s] >       assert (c == 51150)
[   27s] E       assert 54720 == 51150
[   27s] E         +54720
[   27s] E         -51150
[   27s] 
[   27s] dpkt/dpkt.py:218: AssertionError
[   27s] __________________________________ test_utils __________________________________
[   27s] 
[   27s]     def test_utils():
[   27s]         __buf = b'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e'
[   27s]         __hd = '  0000:  00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e     ...............'
[   27s]         h = hexdump(__buf)
[   27s]         assert (h == __hd)
[   27s]         c = in_cksum(__buf)
[   27s] >       assert (c == 51150)
[   27s] E       assert 54720 == 51150
[   27s] E         +54720
[   27s] E         -51150
[   27s] 
[   27s] dpkt/dpkt.py:218: AssertionError
[   27s] ______________________________ test_80211_beacon _______________________________
[   27s] 
[   27s]     def test_80211_beacon():
[...]

PR #510, opened 2020-12-27
Big endian arch bugfixes

obormot:
As was highlighted in #505 some unit tests were failing on IBM System Z, which is a big endian architecture. 
I've reproduced all the failures on QEMU PowerPC VM which is also big endian. In most cases the failures were due to decoders using `socket.ntohs()` to convert from network byte order to host byte order, assuming host is little endian (x86).
On BE systems the function didn't actually convert since the host itself is BE, producing incorrect fields in the decoders.
This patch introduces `compat.ntole()` function which always converts from network order to LE and works consistently on both LE and BE systems.

### DIFF
--- a/dpkt/compat.py
+++ b/dpkt/compat.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-import struct
+from struct import pack, unpack
 import sys
 
 if sys.version_info < (3,):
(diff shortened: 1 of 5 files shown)


dpkt/compat.py
```diff
@@ -37,3 +38,10 @@ def iteritems(d, **kw):
 
     # python3 will return an int if you round to 0 decimal places
     intround = round
+
```
brifordwylie, line 41: I like this... Explicit = good

brifordwylie (approved):
👍

Issue #529, opened 2021-01-21
Test precision issues


brifordwylie:
So looks like some of the new tests have precision issues (when testing on **Mac OSX**).

@crocogorical can you look at these?

```
dpkt/asn1.py::test_utctime FAILED
dpkt/asn1.py::test_decode FAILED


FAILURES 
__________________________________________________ test_utctime ___________________________________________________

    def test_utctime():
        buf = (
            '201005'  # yymndd
            '012345'  # hhmmss
            '+1234'   # +hhmm
        )
>       assert utctime(buf) == 1601815785.0
E       AssertionError: assert 1601840985.0 == 1601815785.0
E        +  where 1601840985.0 = utctime('201005012345+1234')

dpkt/asn1.py:216: AssertionError
___________________________________________________ test_decode ___________________________________________________

    def test_decode():
        import pytest
        from binascii import unhexlify
        buf = unhexlify(
            '20'   # CONSTRUCTED
            '80'   # 128 | 0
        )
        assert decode(buf) == [(32, []), (32, [])]

        # unpacking UTC_TIME
        buf = unhexlify(
            '17'  # t: code: UTC_TIME
            '81'  # l_: code: 128 | 1 (constructed
            '22'  # data len
            '3230313030353031323334352b30303030'

        )
>       assert decode(buf) == [(23, 1601861025.0)]
E       assert [(23, 1601886225.0)] == [(23, 1601861025.0)]
E         At index 0 diff: (23, 1601886225.0) != (23, 1601861025.0)
E         Full diff:
E         - [(23, 1601886225.0)]
E         ?            - ^
E         + [(23, 1601861025.0)]
E         ?             ^^
```

crocogorical:
Interesting! I'm doing all my development on macOS! What versions of python and macOS are you seeing this in?

crocogorical:
Ahha.. this is dumb. This method ultimately ends up calling `time.mktime`, which returns the value in local time, not UTC. Submitted PR with it actually returning the time in UTC.

brifordwylie:
Closed with PR #531

PR #530, opened 2021-01-22
Add test to catch the case where UDP checksum is calculated to be zero.

crocogorical:
It took me a while to come up with the correct parameters to use to hit this line, but this test executes the line in `ip.py` which will modify the UDP checksum to all ones in the case that it has been calculated to be all zeroes.

This takes `ip.py` to 100% test coverage.


brifordwylie:
@crocogorical looks good, merged. Hey can you take at peak at this issue? https://github.com/kbandla/dpkt/issues/529

PR #531, opened 2021-01-22
asn1.py:utctime method now returns time in UTC, instead of local.

crocogorical:
Resolving #529.


brifordwylie:
Looks good on my Mac

```
  py27: commands succeeded
  py35: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
  docs: commands succeeded
  congratulations :)
```

PR #532, opened 2021-01-23
Update radiotap attribute setters.

crocogorical:
Previous implementation had inconsistency between getters (expecting array of bytes), and setters (expecting an int), which meant all setters would fail.

All previous tests pass, and new ones test the getters/setters.


brifordwylie:
Looks good to me...shrug.. not really familiar with radiotap but the changes look solid.

PR #534, opened 2021-01-23
Testing ssl

crocogorical:
Extending testing of `ssl.py` to cover SSL2 class. Note that the `msg` and `pad` elements of `__hdr__` have been removed as the superclass unpacking would extract a single element from the buffer, setting `.data` incorrectly.

They are now only created during the class's `unpack` method.

Other tests are to exercise the exception handling in class unpackings.


brifordwylie:
This looks good @crocogorical, I'd like to have @obormot review it before we merge. :)

PR #538, opened 2021-01-25
Fix/GRE packing byteorder

crocogorical:
GRE class previously unpacked its optional headers with network byte order, but then repacked them with an unspecified order (Line 117). In this case, struct uses the native byte order.

This would result in a byte order swapping for anyone on a little-endian machine who unpacks and repacks a GRE class with optional headers.

Corrected previous test which assumed implementation was correct, and extended testing too.


crocogorical:
@brifordwylie was this left off of the previous merges deliberately? Anything I need to update?

PR #539, opened 2021-01-25
Extend ethernet testing

crocogorical:
Implement a test case which will decode the next layer as Novell "raw" 802.3, and demonstrate that packing with more than 2x VLAN tags raises an exception.

### DIFF
--- a/dpkt/ethernet.py
+++ b/dpkt/ethernet.py
@@ -405,7 +405,6 @@ class VLANtagISL(dpkt.Packet):
 
 
 def test_eth():
-    from . import ip  # IPv6 needs this to build its protocol stack
     from . import ip6
     from . import tcp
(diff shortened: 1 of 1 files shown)


dpkt/ethernet.py
```diff
@@ -667,7 +680,6 @@ def test_eth_llc_ipx():  # 802.3 Ethernet - LLC - IPX
 
 
 def test_eth_pppoe():   # Eth - PPPoE - IPv6 - UDP - DHCP6
-    from . import ip  # IPv6 needs this to build its protocol stack
```
obormot, line 670: I know it's an unused import but I had specifically left it there 'cause otherwise the test would fail (hence the comment). Fine to remove if the renewed test works.

crocogorical: I saw the comment, but wasn't able to induce any failures in py27, py35-py39, or pypy3.7.9 during testing. If there's another way of testing that will fail, please let me know!

obormot: no failures is good!

PR #547, opened 2021-01-31
Remove unreachable exception handling in ip6.py

crocogorical:
During packing of IP6 instances, it does not appear to be possible to raise an AttributeError during this section of code.

 - The `if` statement on line 117 already reads `self.data.sum`
 - No attributes are accessed by `dpkt.in_cksum_done`.

It is possible that `self.data.sum` could be read-only, which would raise an `AttributeError` when trying to write to it.

I cannot think of a way to generate this exception, so we should remove the code, or design a test for it.


brifordwylie (approved):
Your explanation makes sense.. so the PR LGTM

brifordwylie:
@crocogorical Feel free to merge at your convenience

PR #550, opened 2021-02-06
Implement coveralls as Github Action

crocogorical:
An option to address lack of code coverage after deprecating Travis.

This improves on the previous implementation by combining the coverage reports from all of the test runs into a single report. This means that we have 100% coverage of compat.py as the coverage report from python2.7 is also included.

There's a bug somewhere that means that only the python 3.8 run recognises property function definitions as having been run, despite the function body having coverage (example: https://coveralls.io/jobs/74694339/source_files/5716953912)

Coveralls currently reports the aggregated coverage as 99.812%, which is pretty nice.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -23,16 +23,37 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
+
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         python -m pip install flake8 pytest coverage pytest-cov
+
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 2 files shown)


brifordwylie:
Looks great, please go ahead and merge.

crocogorical:
Explanation for the lack of coverage of the function definitions in python3.7-, from the [coverage.py FAQ](https://coverage.readthedocs.io/en/coverage-5.4/faq.html#frequently-asked-questions):
```
Q: My decorator lines are marked as covered, but the “def” line is not. Why?

Different versions of Python report execution on different lines. Coverage.py adapts its behavior to the version of Python being used. In Python 3.7 and earlier, a decorated function definition only reported the decorator as executed. In Python 3.8 and later, both the decorator and the “def” are reported. If you collect execution data on Python 3.7, and then run coverage reports on Python 3.8, there will be a discrepancy.
```

Problem is not our end, so nothing to do.

PR #554, opened 2021-02-07
Extend python versions testing run against.

crocogorical:
Include python 3.9, and also pypy3, as the README says that it's supported, but never explicitly tested against.


crocogorical:
I am not sure why coveralls is reporting the coverage change of the 2.7 run as negative 0, but rather than fix it, I plan to wait until the problem goes away with deprecation of 2.7 support.

PR #562, opened 2021-04-22
fix ipv6 more fragments flag setter

obormot:


dpkt/ip6.py
```diff
@@ -340,13 +340,18 @@ def test_ip6_routing_header():
 
 
 def test_ip6_fragment_header():
-    s = b'\x06\xee\xff\xfb\x00\x00\xff\xff'
+    s = b'\x06\xee\xff\xf9\x00\x00\xff\xff'
```
obormot, line 343: changed the reserved field from 01 to 00 since it doesn't have a setter

PR #582, opened 2021-06-04
added a check in sctp.unpack to prevent exception

K-Reddy1:
when unpack is parsing a malformed sctp packet with improperly labeled chunk size, it will either infinitely loop (if size is labeled as 0), or throw an out of bounds exception (if listed size is larger than the real chunk size)

### DIFF
--- a/dpkt/sctp.py
+++ b/dpkt/sctp.py
@@ -51,7 +51,7 @@ class SCTP(dpkt.Packet):
         while self.data:
             chunk = Chunk(self.data)
             l_.append(chunk)
-            if len(chunk) == 0 or len(chunk) >= len(self.data):
+            if len(chunk) == 0:
                 break
             self.data = self.data[len(chunk):]
         self.chunks = l_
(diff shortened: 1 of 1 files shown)


obormot:
@K-Reddy1 Looks like one of the unit tests is failing. Please take a look

obormot:
@K-Reddy1 any chance you could add a unit test for this condition?
Looks like you already have the example data in the ticket:
```
dd = (
   b'\x00\x00\x00\x01\x00\x06\x00\x1b\x17\x05\x9e\x1c\x00\x00\x08\x00'
   b'\x45\x00\x00\x34\xcc\x09\x40\x00\x1e\x11\x41\x54\x95\xab\x7e\x01'
   b'\x3b\xa6\x00\x09\xb7\x51\x26\xab\x00\x20\xf3\xe8\x32\xbd\xab\xae'
   b'\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
   b'\x00\x00\x00\x00'
)
```
I'm Ok to merge this as is, but it'd be nice to not lose 100% test coverage on the parser.

K-Reddy1:
@obormot sounds good, unit test is on the way

PR #606, opened 2021-08-13
Misc cleanup

brifordwylie:


obormot:
@brifordwylie any clues to why Coveralls show decreased coverage whereas the change is 0.0%? (like here) 
When opening the report there's no decrease in coverage. Feels like there's a condition somewhere that says 
```
if change > 0      # but it should be >=
  green
else
  red
```

obormot (approved):

brifordwylie:
@obormot @kbandla We can change the 'decrease threshold' (see https://github.com/MapServer/MapServer/issues/6223) I don't have access so @kbandla will have to do it... (go here https://coveralls.io/github/kbandla/dpkt/settings)

PR #615, opened 2021-10-07
fix endianness issues in unit tests

obormot:


obormot:
> Hmm, I have apparently missed something yet. With [this patch](https://build.opensuse.org/package/view_file/devel:languages:python/python-dpkt/skip-BE-tests.patch?expand=1) on 1.9.7.2 tarball, I get:
> 
> ```
> [   77s] ____________________________ test_custom_read_write ____________________________
[....]
> [   77s]         # .valid_pcapng buf was collected on a little endian system
> [   77s] >       assert fobj.getvalue() == buf
> [   77s] E       assert b"\n\r\r\n\x0...0\x00\x00\x84" == b"\n\r\r\n|\x...4\x00\x00\x00"
[....]
> ```
> 
> Fortunately, it is just the only test which fails now on s390x. [Full build log](https://github.com/kbandla/dpkt/files/7309145/_log.txt) illustrates all packages used and steps taken.
> 
> What I am missing?

Looks like the diff wasn't fully applied?  The assert above (`assert fobj.getvalue() == buf`) shouldn't have been run on the big endian system.  I've added an if condition there but it's missing in the pytest output above

mcepl:
> This: https://github.com/kbandla/dpkt/pull/615/files#diff-b7ac13862bd7e7717d0e8da77c39e82c27cd64af621798741ec74ae3df39c9a6R1151

You are right. Thank you, that helped.

PR #618, opened 2021-12-02
Add IPv6 support to MPLS.

pajarom:
Very small change to add IPv6 support to MPLS following the same approach used for IPv4.

### DIFF
--- a/dpkt/ethernet.py
+++ b/dpkt/ethernet.py
@@ -118,7 +118,7 @@ class Ethernet(dpkt.Packet):
             if compat_ord(buf[0]) == 0x45:  # IP version 4 + header len 20 bytes
                 self._next_type = ETH_TYPE_IP
 
-            elif ( compat_ord(buf[0]) & 0xf0 ) >> 4 == 6:  # IP version 6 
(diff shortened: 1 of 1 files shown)


dpkt/ethernet.py
```diff
@@ -118,6 +118,9 @@ def _unpack_data(self, buf):
             if compat_ord(buf[0]) == 0x45:  # IP version 4 + header len 20 bytes
                 self._next_type = ETH_TYPE_IP
 
+            elif ( compat_ord(buf[0]) & 0xf0 ) >> 4 == 6:  # IP version 6
```
obormot, line 121: please remove extra whitespace inside parens ```elif (compat_ord(buf[0]) & 0xf0) >> 4 == 6:```
or refactor to ```compat_ord(buf[0]) & 0xf0 == 0x60``` which would be a more compact check

other than this minor comment the change looks good.  
Thanks for the PR!

pajarom: Nice catch. Thanks!

obormot (approved):

Issue #622, opened 2022-02-05
Extract Sequence Num & Fragment Num from 802.11 Pkt


namyamalik:
How can I extract sequence number and fragment number of a 802.11 packet using dpkt? I don't see these fields in the source code for dpkt.ieee80211 (https://dpkt.readthedocs.io/en/latest/_modules/dpkt/ieee80211.html).

I am able to parse the 802.11 packets to extract other info such as src_mac, dst_mac, pkt_type, etc. Just have not been able to get sequence & fragment number.

PR #628, opened 2022-04-22
Fixed Beacon byte ordering.

rpcope1:
When using dpkt, I noticed that the Beacon intervals, timestamps, and beacon capability field were being decoded in the wrong byte order. This change should address this, and adds a test to verify some basic level of correctness.

### DIFF
--- a/dpkt/ieee80211.py
+++ b/dpkt/ieee80211.py
@@ -468,6 +468,7 @@ class IEEE80211(dpkt.Packet):
         )
 
     class Beacon(dpkt.Packet):
+        __byte_order__ = "<"
         __hdr__ = (
             ('timestamp', 'Q', 0),
             ('interval', 'H', 0),
@@ -988,3 +989,11 @@ def test_action_unpack():
     )
     with pytest.raises(dpkt.UnpackError, match="KeyError: category=1 code=0"):
         IEEE80211.Action(buf)
+
+
+def test_beacon_unpack():
+    beacon_payload = b"\xb9\x71\xfa\x45\x52\x02\x00\x00\x64\x00\x11\x04"
+    beacon = IEEE80211.Beacon(beacon_payload)
+    assert beacon.timestamp == 0x0000025245fa71b9
+    assert beacon.interval == 100
+    assert beacon.capability == 0x0411


rpcope1:
After thinking about other possible byte ordering issues for 802.11, it looks like the fraq_seq field is also decoded in the wrong byte order, though I have no idea what the implications of altering (fixing?) that may be for downstream users. I've added
some extra code after looking at #622 to allow easy extraction of the correct fragment and sequence numbers, where applicable.

dpkt/ieee80211.py
```diff
@@ -740,7 +759,7 @@ def test_80211_data():
 def test_80211_data_qos():
     s = (
         b'\x88\x01\x3a\x01\x00\x26\xcb\x17\x44\xf0\x00\x23\xdf\xc9\xc0\x93\x00\x26\xcb\x17\x44\xf0'
-        b'\x20\x7b\x00\x00\xaa\xaa\x03\x00\x00\x00\x88\x8e\x01\x00\x00\x74\x02\x02\x00\x74\x19\x80'
+        b'\x21\x7b\x00\x00\xaa\xaa\x03\x00\x00\x00\x88\x8e\x01\x00\x00\x74\x02\x02\x00\x74\x19\x80'
```
obormot, line 762: could you explain this change?  unit test data usually comes from pcaps so assuming the input is correct, why change it?

dpkt/ieee80211.py
```diff
@@ -474,6 +486,11 @@ class Beacon(dpkt.Packet):
             ('capability', 'H', 0)
         )
 
+        def unpack(self, buf):
+            dpkt.Packet.unpack(self, buf)
+            self.timestamp = struct.unpack('<Q', struct.pack('!Q', self.timestamp))[0]
```
obormot, line 491: So the `compat.ntole` function does something very similar with a 2-byte word..   I suggest converting this into a similar function under the compat module. 
```
def ntole(v):
    """convert a 2-byte word from the network byte order (big endian) to little endian;
    replaces socket.ntohs() to work on both little and big endian architectures
    """
    return unpack('<H', pack('!H', v))[0]
```

dpkt/ieee80211.py
rpcope1: That's fair. I sort of mindlessly altered that to have a test with a non-zero fragment number. I've reverted that alteration, and added an additional unit test for verifying fragment and sequence number behavior.

dpkt/ieee80211.py
rpcope1: Done, I've added this as ntole64(v)

rpcope1:
I guess the question I have at this point, is that my reading of the 802.11 specification seems to indicate that most of the fields are network byte order. It looks like a lot of the parsing code is unpacking the data as little endian. I think that might be worth addressing from a correctness standpoint, but that also may break existing users that expect the current behavior.  Is there a process or procedure for handling this?

obormot:
> I guess the question I have at this point, is that my reading of the 802.11 specification seems to indicate that most of the fields are network byte order. It looks like a lot of the parsing code is unpacking the data as little endian. I think that might be worth addressing from a correctness standpoint, but that also may break existing users that expect the current behavior. Is there a process or procedure for handling this?

I'm going to merge this PR as the work here looks complete. Suggesting we move this question into a new issue @rpcope1 . Thanks!

PR #630, opened 2022-06-02
Support for PCAPNG Packet Block

Schwaggot:
Although the PCAPNG Packet Block type was deprecated a while ago, there are still capture files out there that are written in this format, cf. https://www.ietf.org/staging/draft-tuexen-opsawg-pcapng-02.html#name-packet-block-obsolete

When merged this will add simple support for parsing Packet Block types from PCAPNG files. The type definition was already there, so I guess support was planned anyways?


obormot:
Please add unit tests

Schwaggot:
Added unit tests.

PR #631, opened 2022-06-02
Graceful handling of PCAPNG option comment UTF-8 decoding errors

Schwaggot:
I believe before the current PCAPNG standard it was customary to have zero-terminated C-strings as comments, similar to the PCAP format. Although by now deprecated and switched to UTF-8 strings there are still capture files out there that have this old format. At the moment dpkt would just crash with a `UnicodeDecodeError` while trying to read these comments, which is a bit harsh for a field that has usually no contribution to the capture itself.

When merged this will add gracefull support for UTF-8 decoding errors, i.e., it will first try to parse comments as UTF-8 (new standard), if that fails it will check whether the data contains a zero-terminator and interpret it as a C-string. If no zero-terminator is found, the `UnicodeDecodeError` will still be raised (as previously).

### DIFF
--- a/docs/authors.md
+++ b/docs/authors.md
@@ -74,6 +74,6 @@ Dug Song \<<dugsong@monkey.org>\>
     Extend test coverage
 
   - Schwaggot \<<tom.ryan@posteo.de>\>  
-    Graceful handling of PCAPNG option comment UTF-8 decoding errors
+    pcapng module, support for Packet Block and handling of UTF-8 option comments
 
 If you want to contribute to dpkt, see `contributing`.
--- a/docs/authors.md
+++ b/docs/authors.md
@@ -73,4 +73,7 @@ Dug Song \<<dugsong@monkey.org>\>
   - crocogorical \<<crocogorical@gmail.com>\>  
     Extend test coverage
 
+  - Schwaggot \<<tom.ryan@posteo.de>\>  
+    Graceful handling of PCAPNG option comment UTF-8 decoding errors
+
 If you want to contribute to dpkt, see `contributing`.
--- a/dpkt/pcapng.py
+++ b/dpkt/pcapng.py
@@ -194,7 +194,7 @@ class PcapngOption(dpkt.Packet):
                 self.text = self.data.decode('utf-8')
             except UnicodeDecodeError as ude:
                 if b'\x00' in self.data:
-                    self.text = self.data[0:self.data.index(b'\x00')]
+                    self.text = self.data[0:self.data.index(b'\x00')].decode('ascii')
                 else:
                     raise ude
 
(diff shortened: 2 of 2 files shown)


Schwaggot:
Added unit tests.

dpkt/pcapng.py
```diff
@@ -759,6 +765,50 @@ def test_epb():
...
+        b'\x06\x00\x00\x00\x7c\x00\x00\x00\x01\x00\x00\x00\xff\xff\xff\xff\x79\xd2\xdf\xe1\x44\x00'
+        b'\x00\x00\x44\x00\x00\x00\x00\x00\x00\x01\x00\x06\x00\x0b\xdb\x43\xe7\x4b\xf6\x7f\x08\x00'
+        b'\x45\x00\x00\x34\x2b\x1f\x40\x00\x40\x06\x15\x63\x82\xd9\xfa\x81\x82\xd9\xfa\x0d\x17\x70'
+        b'\xec\x3e\x02\xba\x94\x38\x81\x52\x4a\x39\x80\x10\xbb\x5d\x53\x0d\x00\x00\x01\x01\x08\x0a'
+        b'\x03\xf9\xc7\xbf\x04\x02\x38\x28\x01\x00\x0f\x00\x50\x61\x63\x6b\x65\x74\x20\x23\x31\x20'
+        b'\x78\x4d\x39\x87\x0c\x00\x00\x00\x00\x00\x7c\x00\x00\x00')
+
+    try:
+        EnhancedPacketBlockLE(buf)
+        assert False
```
Schwaggot, line 807: Need guidance on this test. It currently lowers the code coverage because `assert False` is never reached. Not quite sure how to create a test that expects an exception and fails otherwise that reaches 100% code coverage.

obormot: `assert False` doesn't test any code, so I'd rather just remove it.

docs/authors.md
```diff
@@ -73,4 +73,7 @@ Dug Song \<<dugsong@monkey.org>\>
   - crocogorical \<<crocogorical@gmail.com>\>  
     Extend test coverage
 
+  - Schwaggot \<<tom.ryan@posteo.de>\>  
+    Graceful handling of PCAPNG option comment UTF-8 decoding errors
```
obormot, line 77: since both of your PRs are in pcapng.py, I suggest merging the 2 changes to `authors.md` together.  can be done after one of the PRs is merged

dpkt/pcapng.py
Schwaggot: Alright, removed.

docs/authors.md
Schwaggot: Adjusted, both have same text, so it shouldn't matter in which order they are merged.

PR #635, opened 2022-06-24
Create bfcp.py

amoxuk:
add bfcp

### DIFF
--- a/dpkt/bfcp.py
+++ b/dpkt/bfcp.py
@@ -1,0 +1,744 @@
+# $Id: bfcp.py amoxuk $
+# -*- coding: utf-8 -*-
+import struct
+from binascii import unhexlify
+
+import dpkt
+
+
+######################
+#     Primitive      #
+######################
+class Primitive:
+    FloorRequest = 1
+    FloorRelease = 2
+    FloorRequestQuery = 3
+    FloorRequestStatus = 4
+    UserQuery = 5
+    UserStatus = 6
+    FloorQuery = 7
+    FloorStatus = 8
+    ChairAction = 9
+    ChairActionAck = 10
+    Hello = 11
+    HelloAck = 12
+    Error = 13
+    FloorRequestStatusAck = 14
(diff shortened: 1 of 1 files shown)


dpkt/bfcp.py
```diff
@@ -0,0 +1,744 @@
...
+        elif isinstance(self.attrs, BFCPAttr):
+            data += bytes(self.attrs)
+        self.len = int(len(data) / 4)
+        return self.pack_hdr() + data
+
+    def unpack(self, buf):
+        dpkt.Packet.unpack(self, buf)
+        l_ = []
+        if self.f:
+            pass
```
obormot, line 219: ?

dpkt/bfcp.py
```diff
@@ -0,0 +1,744 @@
...
+            if isinstance(self.value, list):
+                data += struct.pack('!H', self.value[0])
+                data += b''.join(bytes(attr) for attr in self.value[1:])
+            else:
+                data += struct.pack('!H', self.value)
+        elif self.type in (AttrType.SUPPORTED_PRIMITIVES, AttrType.SUPPORTED_ATTRIBUTES, AttrType.REQUEST_STATUS,
+                           AttrType.PRIORITY):
+            if AttrType.SUPPORTED_ATTRIBUTES == self.type:
+                self.value = [d << 1 for d in self.value]
+            data += struct.pack(f'!{len(self.value)}B', *self.value)
```
obormot, line 305: build failed because the f-strings are not supported in older versions of python

dpkt/bfcp.py
amoxuk: unimplemented feature...
[The Fragmentation (F) flag](https://www.rfc-editor.org/rfc/rfc8855.html#name-common-header-format-2)

obormot: adding this as a comment in code would be helpful

amoxuk: got it.

PR #638, opened 2022-08-30
Make example code python3-compatible

diohabara:
## What

Rewrite docs to make them compatible with Python3. The current versions are following Python2.

## Why

The source code are already using Python3, and the need for Python3 is seemingly bigger.


obormot:
Thanks for your contribution @diohabara !

Issue #462, opened 2019-11-19
Feature Request: Decimal timestamps all the time


gmacon:
Currently, `dpkt.pcap.Reader` returns the timestamp as a `float` if the pcap has microsecond resolution and as a `decimal.Decimal` if the pcap has nanosecond precision. I'd like to request that the timestamps be represented as `Decimal` objects all the time.

I realize that this will be a backwards-incompatible change, so I'd be OK with this behavior being opt-in for now, but I'd like to see it become the default behavior at some point.

PR #639, opened 2022-09-16
Use int precision multiplier instead of float

andrew-bolin:
This code change is very straightforward (`1e6` -> `1_000_000`), but I'll explain the backstory.

I have a software utility that writes to either pcap or pcapng - I'm sure many people do, the API is the same which makes it very easy :D 

Except! We have nanosecond resolution timestamps. So we're using nano=True for pcap (not supported in pcapng, that's fine).

It turns out that floats do not have sufficient precision to encode a UNIX timestamp with small nanosecond differences, so we use `Decimal` for our timestamp. 
This is all well and good for pcap files (which multiply the timestamp by an integer), but no good for pcap, which multiplies the timestamp by 1e6 (which is a float).

So, in my project, I _could_ do something like `if pcapng, convert timestamp to float`.

But it's a lot cleaner if dpkt could replace the constant `1e6` with `1_000_000`. Then I can simply pass a Decimal into either Writer type :sunglasses: 

I also converted it from being a magic number in 4 places to a variable, this might help if someone in the future feels like expanding the pcapng Writer to support nanoseconds (if the file format permits?)... :wink:

### DIFF
--- a/dpkt/pcapng.py
+++ b/dpkt/pcapng.py
@@ -608,7 +608,7 @@ class Reader(object):
             raise ValueError('IDB not found')
 
         # set timestamp resolution and offset
-        self._divisor = float(1e6)  # defaults
+        self._divisor = 1000000  # defaults
         self._tsoffset = 0
         for opt in idb.opts:
             if opt.code == PCAPNG_OPT_IF_TSRESOL:
@@ -616,7 +616,7 @@ class Reader(object):
                 # if MSB=1, the remaining bits is a neg power of 2 (e.g. 10 means 1/1024 of second)
                 opt_val = struct_unpack('b', opt.data)[0]
                 pow_num = 2 if opt_val & 0b10000000 else 10
-                self._divisor = float(pow_num ** (opt_val & 0b01111111))
+                self._divisor = pow_num ** (opt_val & 0b01111111)
 
             elif opt.code == PCAPNG_OPT_IF_TSOFFSET:
                 # 64-bit int that specifies an offset (in seconds) that must be added to the
--- a/dpkt/pcapng.py
+++ b/dpkt/pcapng.py
@@ -421,6 +421,7 @@ class Writer(object):
         idb can be an instance of InterfaceDescriptionBlock(LE) (or sequence of them)
         """
         self.__f = fileobj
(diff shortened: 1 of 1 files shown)


obormot:
Build fails due to Python 2.7 not understanding the 1_000_000 convention. dpkt will drop python2 support any day now but for now replacing 1_000_000 with 1000000 will help the build pass

obormot:
This PR changes the Writer class adding `self._precision_multiplier = 1000000`, replacing the hardcoded value of `1e6`; however the existing Reader class already has `self._divisor = float(1e6)` so at a minimum this change would introduce inconsistency between the Reader and the Writer, which we should avoid.

obormot:
Also https://github.com/kbandla/dpkt/issues/462 could be related here?

andrew-bolin:
Fair point!

My change makes the two Writers more consistent, but I didn't think to look
at the Reader.

Next time I'm in front of a computer I'll swap an integer constant into the
Reader as well.

Regarding issue #462 - indirectly related, yes. If one were to return
Decimal timestamps as suggested there, it would reasonably be expected to
accept them as inputs also, which is the goal of this change 🙂

andrew-bolin:
oh, Python2 treats `int / int` as an integer division, unlike Python3 :cry: 

how far away is that plan to drop 2.x ?

andrew-bolin:
@obormot - anything else I can do to help this proceed?

obormot:
> @obormot - anything else I can do to help this proceed?

TBH I'd like to spend some more time reviewing this change, not code-wise but conceptually. I'd like to understand (1) if this could potentially be a breaking change, (2) what's the deal with timestamp precision in pcaps, should they be ints or floats, or Decimal as some other PRs suggested -- essentially what is the "right way" to do it.

obormot:
> oh, Python2 treats `int / int` as an integer division, unlike Python3 😢

Ah, this must be the reason we used floats in the original code (written for py2 at the time). 
I've reviewed this change once more and it looks good. Merging into master for now, let's monitor and see if anyone complains :) 
In general though there are discrepancies between pcap and pcapng handling (lack of nano=True option in pcapng, use of Decimal and 1E6 in pcap) which we should address.

obormot (approved):

PR #640, opened 2022-09-26
Fix tls records unpacking

obormot:
Refactor unpacking of TLS. 
`class TLS` is just a container for `TLSRecord`s.  As such, it doesn't need its own `__hdr__`. In current code it just copies the header of the 1st `TLSRecord`, which is technically incorrect. 
A better pattern would be that `TLS` delegates unpacking to the underlying `TLSRecord` and advances the buffer. The proposed change does just that.


obormot:
This is potentially a breaking change, since it removes `TLS.__hdr__`, so going to merge into master and give it some soak time.

PR #645, opened 2022-11-09
Test for and fix implicit usage of native sized structs 

### LINKED ISSUE #644: bitfield unittest might be platform dependant
This test gives me trouble.

https://github.com/kbandla/dpkt/blob/2d30526422d48f83d4f724221d9ae0dcd5ff57fe/dpkt/dpkt.py#L62-L65

I think it should not be called on `ph_struct` but a concat of the classes `__byte_order__` and it, no?

On my system the test tells me `L` is expected to be 8*8bits instead of 32, I'd expect.
https://docs.python.org/3/library/struct.html#format-characters

Am I just overlooking something? I'll provide a PR with a unittest with what I mean tomorrow, if wanted.

AiyionPrime:
The `struct.calcsize`[^1] in line 63 defaults to machine’s native format [^2], as long as not differently specified.

The data one parses should not yield different results based on the parsing machines architecture; which is acknowledged in

https://github.com/kbandla/dpkt/blob/2d30526422d48f83d4f724221d9ae0dcd5ff57fe/dpkt/dpkt.py#L45-L46

for `__hdr_fmt__` but is lacking for `__bit_fields__`.

resolves #644 

[^1]: https://docs.python.org/3/library/struct.html#struct.calcsize
[^2]: https://docs.python.org/3/library/struct.html#byte-order-size-and-alignment

### DIFF
--- a/dpkt/dpkt.py
+++ b/dpkt/dpkt.py
@@ -36,13 +36,14 @@ class PackError(Error):
 class _MetaPacket(type):
     def __new__(cls, clsname, clsbases, clsdict):
         t = type.__new__(cls, clsname, clsbases, clsdict)
+        byte_order = getattr(t, '__byte_order__', '>')
         st = getattr(t, '__hdr__', None)
         if st is not None:
             # XXX - __slots__ only created in __new__()
             clsdict['__slots__'] = [x[0] for x in st] + ['data']
             t = type.__new__(cls, clsname, clsbases, clsdict)
             t.__hdr_fields__ = [x[0] for x in st]
-            t.__hdr_fmt__ = getattr(t, '__byte_order__', '>') + ''.join([x[1] for x in st])
+            t.__hdr_fmt__ = byte_order + ''.join([x[1] for x in st])
             t.__hdr_len__ = struct.calcsize(t.__hdr_fmt__)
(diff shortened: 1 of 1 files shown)


dpkt/dpkt.py
```diff
@@ -36,13 +36,14 @@ class PackError(Error):
...
         t = type.__new__(cls, clsname, clsbases, clsdict)
+        byte_order = getattr(t, '__byte_order__', '>')
         st = getattr(t, '__hdr__', None)
         if st is not None:
             # XXX - __slots__ only created in __new__()
             clsdict['__slots__'] = [x[0] for x in st] + ['data']
             t = type.__new__(cls, clsname, clsbases, clsdict)
             t.__hdr_fields__ = [x[0] for x in st]
-            t.__hdr_fmt__ = getattr(t, '__byte_order__', '>') + ''.join([x[1] for x in st])
+            t.__hdr_fmt__ = byte_order + ''.join([x[1] for x in st])
```
obormot, line 46: (minor) looks like the join can be refactored to omit the square brackets `''.join(x[1] for x in st)`

AiyionPrime: I have not recognized generator expressions to be valid function argument before, thanks.
Will update this friday morning.

PR #654, opened 2023-06-15
Extend python versions testing runs against

AiyionPrime:
> - 3.10: released on 2021-10-04
> - 3.11: released on 2022-10-24

Similar to #554 two years ago.


obormot:
Suggest adding lines to `setup.py` similarly to the https://github.com/kbandla/dpkt/pull/554 referenced

kbandla:
Is this a good time to remove 2.7?

AiyionPrime:
Imho any time is.
I read the question as to change the PR in order to remove 2.7 from the supported versions, right?

AiyionPrime:
Mhm. And the tox file has some version references I missed as well.
Update tomorrow morning.

AiyionPrime:
I think that's it.

The previous updater missed 3.9 in `tox.ini`, I added that as well.
I dropped 2.7 as suggested above. Let me know if there's something missing.

obormot:
> @obormot Any idea, why `build (2.7)` ist still an expected step even after removing it from the workflow yml?

no clue, but try changing run-on: ubuntu-20.04 to ubuntu-latest (which doesn't include python2), and see if it helps?

AiyionPrime:
@obormot Python 3.5 is missing on Ubuntu20 as well.

To be specific:
Python 3.5 was sunsetted 2020-09-30
Python 3.6 was sunsetted 2021-12-23
and
Python 3.7 will be sunsetted in six days. (2023-06-27)

I'd suggest getting rid of already sunsetted versions.

obormot:
I think the reason 2.7, 3.5 and 3.6 builds are still "expected", may be branch protection rules, which are defined in the repo settings (available to @kbandla). Reference: https://github.com/orgs/community/discussions/26698

AiyionPrime:
pypy defaultet to the (unsupported) version 3.6 as per setup-python action v2.
I updated that to the latest v4, updated checkout to v3, as v4 recommends it, and set pypy to 3.7 explicitely.

We always tested for one pypy version and not all supported.
I assume that's intended?

If we want to test for more let me know and I'll update this PR accordingly.
Other than that the only apparent Issue I see is `2.7`, `3.5` and `3.6` to be expected.

After @kbandla resolved that this might be good to go.

kbandla:
updated the branch protection settings for master (removed 2.7-3.5). I will add another rule for the 2.7 branch, when we freeze it.

AiyionPrime:
Then this is ready for review;
only question remaining is the lack of additional pypy versions, which might be intended.

obormot:
This looks good to me.  Not sure about pypy, never used it.  We can update the README separately once the 2.7 support branch is up.

PR #655, opened 2023-08-09
Fix poll & precision data types

paulgear:
Poll and precision are signed, not unsigned. https://datatracker.ietf.org/doc/html/rfc5905#page-21


paulgear:
@kbandla Merged latest `kbandla:master`.

PR #680, opened 2026-04-19
Add CoAP and LWM2M TLV support

joakimeriksson:
Implement CoAP (RFC 7252) with support for core features, Observe (RFC 7641), and Block-wise transfers (RFC 7959).
Implement LWM2M Binary TLV + JSON (OMA LWM2M v1.0) with support for Object Instances, Multiple Resources, and Resource Instances.
Add functional examples for LWM2M devices and servers, including a Leshan-ready agent.

### DIFF
--- a/dpkt/lwm2m.py
+++ b/dpkt/lwm2m.py
@@ -36,153 +36,345 @@ LWM2M_DEV_UTC_OFFSET = 14
 LWM2M_DEV_TIMEZONE = 15
 LWM2M_DEV_SUPPORTED_BINDING = 16
 
-# LWM2M TLV Type IDs
-LWM2M_TLV_OBJECT_INSTANCE = 0x00  # 00
-LWM2M_TLV_RESOURCE_INSTANCE = 0x40  # 01
-LWM2M_TLV_MULTIPLE_RESOURCE = 0x80  # 10
-LWM2M_TLV_RESOURCE = 0xC0  # 11
+# TLV Type IDs (top two bits of the TLV type byte)
+LWM2M_TLV_OBJECT_INSTANCE = 0x00     # 00
+LWM2M_TLV_RESOURCE_INSTANCE = 0x40   # 01
+LWM2M_TLV_MULTIPLE_RESOURCE = 0x80   # 10
+LWM2M_TLV_RESOURCE = 0xC0            # 11
 
 
-class LWM2M_TLV(dpkt.Packet):
-    """LWM2M Binary TLV Format.
+class LWM2M(object):
+    """LWM2M Protocol namespace (OMA Lightweight M2M).
 
-    OMA LWM2M v1.0, Section 6.3.3.
+    LwM2M payloads ride inside CoAP, so this class is a logical
+    namespace rather than a wire-level packet. It groups:
 
-    Attributes:
-        type (int): TLV Type (Object Instance, Resource, etc.)
-        id (int): Identifier (8 or 16 bits)
-        value (bytes): Value
+    * an abstract data model — :class:`LWM2M.Resource`,
+      :class:`LWM2M.ObjectInstance`
+    * supported wire formats — :class:`LWM2M.TLV` (more to come)
+    * format dispatch — :meth:`LWM2M.encode`, :meth:`LWM2M.decode`
+
+    The data model lets callers construct an LwM2M payload once and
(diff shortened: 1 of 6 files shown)


examples/leshan_agent.py
```diff
@@ -0,0 +1,165 @@
...
+    if len(val) == 4: return struct.unpack('>i', val)[0]
+    if len(val) == 8: return struct.unpack('>q', val)[0]
+    return 0
+
+def encode_float(val):
+    """Encode float to LWM2M binary format (4 or 8 bytes IEEE 754)."""
+    return struct.pack('>f', val) # 4-byte float
+
+def run_leshan_agent():
+    SERVER_IP = '23.97.187.154'
```
kbandla: Should we change this to `leshan.eclipseprojects.io` ? `SERVER_IP ` is not used in the code too.

joakimeriksson: Yes, that would make sense as they might change the Server IP. I will fix that (and make sure not to have unused variables).

joakimeriksson: - Switched the Leshan endpoint to leshan.eclipseprojects.io
- Made the local bind host configurable and avoided binding the local server/device examples to 0.0.0.0
- Updated the Leshan agent to bind to the interface the OS would use to reach the Leshan server, so the public demo still works without binding all interfaces

Tested on Leshan:

image

dpkt/lwm2m.py
```diff
@@ -0,0 +1,190 @@
...
+LWM2M_DEV_SUPPORTED_BINDING = 16
+
+# LWM2M TLV Type IDs
+LWM2M_TLV_OBJECT_INSTANCE = 0x00  # 00
+LWM2M_TLV_RESOURCE_INSTANCE = 0x40  # 01
+LWM2M_TLV_MULTIPLE_RESOURCE = 0x80  # 10
+LWM2M_TLV_RESOURCE = 0xC0  # 11
+
+
+class LWM2M_TLV(dpkt.Packet):
```
kbandla: Really minor thing, but could you consider naming the class `LWM2M` instead? It would *keep the format common with other classes. Doing this might also need to change the example files. 
edit: fix typo keeping->keep

joakimeriksson: Yes, that makes sense. Will fix that - and make it possible to add other formats than TLV in the future (there are multiple possible options on the payload).

joakimeriksson: BTW: realized that this class is only for the TLVs so in fact the name should probably be TLV. Should it be its own file then or is it ok if it is in the same file as the lwm2m implementation? The logic is that the TLV class is managing the TLV list of data in a LWM2M/CoAP packet.

kbandla: Lets keep it in its own file for now. Whenever we get to implement v1.1 and later, we can refactor it like in cdp.py. What do you think?

joakimeriksson: I did a "big" update introduced multi-format in LWM2M - tested with Leshan - read out data both via binary TLV and proper JSON format. See below:
```
[*] Sent Registration for dpkt-agent-2026
[*] Received CoAP GET /3 (Accept=11542)
    [+] Responded /3 (format 11542, 147B): 080090c8000c64706b742d70726f6a656374c8010e636c692d6167656e742d76312e32c8020c534e2d323032362d44504b54c503312e392e388306410001840742000ed8c1095fc20a0400c40d69f07b83c60e2b30313a3030c80f104575726f70652f53746f636b686f6c6dc8110e44504b542d4d61702d4167656e74c8120c76312e302d7669727475616cc11401c2150800
[*] Received CoAP GET /3 (Accept=11543)
    [+] Responded /3 (format 11543, 398B): {"bn":"/3/0/","e":[{"n":"0","sv":"dpkt-project"},{"n":"1","sv":"cli-agent-v1.2"},{"n":"2","sv":"SN-2026-DPKT"},{"n":"3","sv":"1.9.8"},{"n":"6/0","v":1},{"n":"7/0","v":3800},{"n":"9","v":95},{"n":"10","v":1024},{"n":"13","v":1777367945},{"n":"14","sv":"+01:00"},{"n":"15","sv":"Europe/Stockholm"},{"n":"17","sv":"DPKT-Map-Agent"},{"n":"18","sv":"v1.0-virtual"},{"n":"20","v":1},{"n":"21","v":2048}]}

```

PR #682, opened 2026-08-10
tftp: raise dpkt error instead of IndexError on a truncated request

eeshsaxena:
### Problem

`TFTP.unpack()` handles read/write requests by splitting the payload on NUL and reading two fields:

```python
if self.opcode in (OP_RRQ, OP_WRQ):
    l_ = self.data.split(b'\x00')
    self.filename = l_[0]
    self.mode = l_[1]
```

A request whose payload contains no NUL separator (a malformed or truncated packet) makes `split` return a single-element list, so `l_[1]` raises a raw `IndexError` rather than a `dpkt.UnpackError`:

```python
import dpkt
dpkt.tftp.TFTP(b'\x00\x01abc')   # IndexError: list index out of range
```

The DATA/ACK/ERR opcodes already surface truncation as a dpkt error, because `Packet.__init__` converts `struct.error` to `NeedData`/`UnpackError`. The RRQ/WRQ path bypasses that because it fails with `IndexError`.

### Fix

Raise `dpkt.NeedData` when the request does not carry both a filename and a mode, keeping malformed TFTP packets within dpkt's own exception hierarchy.

### Tests

Added `test_op_rrq_truncated` and `test_op_wrq_truncated`. Both fail with `IndexError` before the change and pass after; the existing TFTP tests still pass (6 total).


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #683, opened 2026-08-10
Raise dpkt error instead of IndexError on truncated SSL2 and BGP EVPN records

eeshsaxena:
Two more truncated-packet paths (same class as #682) that leak a raw `IndexError` instead of a `dpkt` error.

### SSL2

`SSL2.unpack` reads the padding-length byte in the 3-byte header form with `compat_ord(self.data[0])`. A record that stops after the 2-byte length field leaves `self.data` empty:

```python
import dpkt
dpkt.ssl.SSL2(b'\x00\x01')   # IndexError: index out of range
```

### BGP EVPN

`RouteEVPN.unpack` reads the MAC and IP address length bytes with `compat_ord(buf[0])` without checking the buffer:

```python
dpkt.bgp.RouteEVPN(b'\xff\xff\xff\xff')   # IndexError: index out of range
```

### Fix

Guard each read and raise `dpkt.NeedData`, consistent with the other truncated-packet handling in these modules.

### Tests

Added `test_ssl2_truncated_padding_length` (ssl.py) and `test_route_evpn_truncated` (bgp.py), covering the SSL2 padding byte and both the MAC (type 2) and IP (type != 1) length reads. They fail with `IndexError` before this change and pass after. `ssl.py` and `bgp.py` tests pass (62).


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

